PR #2, opened 2022-04-09
Do not strip directory name if origin dir is empty

wouterj:
If you have a repository dedicated to documentation (e.g. symfony-docs), `$directory` will be set to an empty string (instead of e.g. `docs/`). In these cases, stripping 0 + 1 characters from the start produces non existent file names.

The code should only strip characters if a subdirectory is given.


jaapio:
I closed this PR because I decided to have a mono repo for the guides. I do agree with you that you will need to edit 3 repo's in the same time in many cases. as the repos are highly connected.

wouterj:
I've recreated this PR in https://github.com/phpDocumentor/guides/pull/11

PR #11, opened 2022-04-13
Do not strip directory name if origin dir is empty

wouterj:


jaapio (approved):

jaapio:
Nice thanks for pushing this again!

PR #12, opened 2022-04-13
Remove the package name to avoid using the mono-repo in composer by mistake

stof:


jaapio:
Removing the name is breaking the checks on the composer.json. maybe we should put in some random name like `phpdocumentor/guides-source-repo`?

stof:
if you check the validity of the composer.json file at the root of the mono-repo, you should pass `--no-check-publish` as this composer.json is not meant to be published.

jaapio:
Thanks for the suggestions I implemented this in #59

PR #21, opened 2022-04-28
Add node visitor pattern + implement internal references (the `:ref:` role)

### LINKED ISSUE #16: Add ref role implementation
https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#the-ref-role

Currently, the `DocResolver` claims to be able to handle both `:ref:` and `:doc:` roles, but the implementation only implements `:doc:` roles.

`:ref:` allows you to reference so-called [internal hyperlink targets](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#hyperlink-targets) inside the project (not necessarily the current document). We use this a lot in the Symfony documentation to link users to specific sections of another page.
However, the current ref implementation works like doc: it allows you to link to documents inside the project.

In order to properly support the ref role, the `Metas` class must be refactored a bit. It currently doesn't store internal hyperlink targets in a flat map, which will be required for the `RefResolver`.

If this is not yet part of the roadmap of this package, I'm happy to start the contribution of this feature. I mainly created this issue to see if I'm missing something here and avoid duplicate work.

wouterj:
Fixes #16 

As promised, this PR fixes the `:ref:` role implementation. It is used to link to internal hyperlink targets *within the project* (i.e. can also be used for explicit hyperlink targets in other documents).

Node transformers (visitor pattern)
---

I also added the concept of a "node transformer" (or node visitor), which makes it easy to collect all `AnchorNodes` - so it can be used by the `RefResolver`.

Node transformers is a pattern that Python Sphinx & docutils (rst parser) use a lot: https://github.com/docutils-mirror/docutils/tree/e88c5fb08d5cdfa8b4ac1020dd6f7177778d5990/docutils/transforms They allow to remove more state awareness from the parser/nodes to visitors, e.g. the whole generation of `Metas`/meta entries can be removed from `Parser`  and the nodes into a visitor in the future.
Transformers also allow to add or "collapse" nodes (e.g. create the section nodes based on title nodes, or apply `.. class::` directives on the adjacent node). You can see some more ideas in https://github.com/doctrine/rst-parser/issues/145

For the implementation, I've used Twig's node visitor as a base, as this is a proven pattern in Twig.

Metas
---

I'm not so sure about the role of the `Metas` object in the current guides, and I see there also is a "TODO" item about this in `DocResolver`. In my test project, I've defined `Metas` as a service, which allowed me to directly inject it in the `RefResolver`. This works great in the test project of the Symfony docs. However, the `Parser` somehow resets and creates a new `Metas` instance in some circumstances, breaking this implementation.

Maybe it is a good idea to discuss how `Metas` have to work (also in relation to node transformers) and refactor this part of the Parser as well. cc @jaapio @mvriel

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/MarkupLanguageParser.php
+++ b/packages/guides-restructured-text/src/RestructuredText/MarkupLanguageParser.php
@@ -6,6 +6,7 @@ namespace phpDocumentor\Guides\RestructuredText;
 
 use phpDocumentor\Guides\MarkupLanguageParser as ParserInterface;
 use phpDocumentor\Guides\Nodes\DocumentNode;
+use phpDocumentor\Guides\NodeTransformer\NodeTransformer;
 use phpDocumentor\Guides\ParserContext;
 use phpDocumentor\Guides\RestructuredText\Directives\AdmonitionDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\BestPracticeDirective;
@@ -52,6 +53,9 @@ class MarkupLanguageParser implements ParserInterface
     /** @var Directive[] */
     private $directives = [];
 
+    /** @var iterable<NodeTransformer> */
+    private $transformers;
+
     /** @var string|null */
     private $filename = null;
 
@@ -60,10 +64,14 @@ class MarkupLanguageParser implements ParserInterface
 
     /**
      * @param iterable<Directive> $directives
+     * @param iterable<NodeTransformer> $transformers
      */
     public function __construct(
-        iterable $directives
+        iterable $directives,
+        iterable $transformers = []
     ) {
+        $this->transformers = $transformers;
+
         foreach ($directives as $directive) {
             $this->registerDirective($directive);
         }
@@ -169,6 +177,6 @@ class MarkupLanguageParser implements ParserInterface
 
     private function createDocumentParser(): DocumentParser
     {
-        return new DocumentParser($this, $this->directives);
+        return new DocumentParser($this, $this->directives, $this->transformers);
     }
 }
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/DocumentParser.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/DocumentParser.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace phpDocumentor\Guides\RestructuredText\Parser;
 
 use ArrayObject;
+use phpDocumentor\Guides\NodeTransformer\DocumentNodeTraverser;
+use phpDocumentor\Guides\NodeTransformer\NodeTransformer;
 use phpDocumentor\Guides\Nodes\DocumentNode;
 use phpDocumentor\Guides\Nodes\TitleNode;
 use phpDocumentor\Guides\RestructuredText\Directives\Directive as DirectiveHandler;
@@ -41,18 +43,22 @@ class DocumentParser
     /** @var Productions\Rule */
     private $startingRule;
     private MarkupLanguageParser $parser;
+    private DocumentNodeTraverser $nodeTraverser;
 
     /**
-     * @param DirectiveHandler[] $directives
+     * @param DirectiveHandler[]        $directives
+     * @param iterable<NodeTransformer> $transformers
      */
     public function __construct(
         MarkupLanguageParser $parser,
-        array $directives
+        array $directives,
+        iterable $transformers
     ) {
         $this->documentIterator = new LinesIterator();
         $this->openSectionsAsTitleNodes = new ArrayObject();
 
         $this->startingRule = new Productions\DocumentRule($this, $parser, $directives);
+        $this->nodeTraverser = new DocumentNodeTraverser($transformers);
         $this->parser = $parser;
     }
 
@@ -64,6 +70,8 @@ class DocumentParser
         if ($this->startingRule->applies($this)) {
             $this->startingRule->apply($this->documentIterator, $this->document);
         }
+
+        $this->nodeTraverser->traverse($this->document);
 
         return $this->document;
     }
--- a/packages/guides/src/Metas.php
+++ b/packages/guides/src/Metas.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace phpDocumentor\Guides;
 
 use phpDocumentor\Guides\Meta\Entry;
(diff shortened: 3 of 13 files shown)


jaapio:
Hi Wouter,

Thanks for this great change. It's great to see that you can contribute to this project. I didn't have time yet to look in detail at your code. But I wanted to provide some input on your changes to let you know that I'm still here. 

**Transformers**

I like this addition. In phpDocumentor itself, we call this our compiler step. We are separating this from the parsing step as we want to reflect the parsed content as close as possible in the AST. But also to allow some dynamic imports or circular object references, which are harder to handle when you want to cache the AST. In theory you are doing the same thing, as you added the transformers just before the parser returns the document. If we would move this out of the parser completely. We could introduce a compiling step allowing the transformers to iterate over the full documentation set? So you could even extract information from all documents and manipulate the documents together? An example of such a use case: https://github.com/phpDocumentor/phpDocumentor/blob/master/src/phpDocumentor/Compiler/Pass/TableOfContentsBuilder.php

**Meta's**
I'm not fully aware of all locations meta's are used. But how I understood the logic works is that all meta entries, which are pages, headers, and anchors should be collected in a single meta's object. When running the parser with cache this meta's object is read from the cache. Allowing the parser to process just the files that are not changed. And still, be able to link to all files from the newly parsed files. So basically it works like some kind of table-of-contents for the full documentation. 

I think you are right that there are some situations where this logic breaks. I'm not sure in what situations this is as it was already implemented in some way in the original project. And we tried to keep it as is. I think this happens when a fragment or include is processed. Although I think we can fix this by moving the meta's into the parser context. As the context should be passed around when processing a single file including all content imported into that file. 

I do like your suggestion to move the meta's into the transformation phase. This makes a lot of sense, as it would allow us to remove some more logic from the parser. Making it a more pure parser.

jaapio:
Hi Wouter,

Thanks for this great change. It's great to see that you can contribute to this project. I didn't have time yet to look in detail at your code. But I wanted to provide some input on your changes to let you know that I'm still here. 

**Transformers**

I like this addition. In phpDocumentor itself, we call this our compiler step. We are separating this from the parsing step as we want to reflect the parsed content as close as possible in the AST. But also to allow some dynamic imports or circular object references, which are harder to handle when you want to cache the AST. In theory you are doing the same thing, as you added the transformers just before the parser returns the document. If we would move this out of the parser completely. We could introduce a compiling step allowing the transformers to iterate over the full documentation set? So you could even extract information from all documents and manipulate the documents together? An example of such a use case: https://github.com/phpDocumentor/phpDocumentor/blob/master/src/phpDocumentor/Compiler/Pass/TableOfContentsBuilder.php

**Meta's**
I'm not fully aware of all locations meta's are used. But how I understood the logic works is that all meta entries, which are pages, headers, and anchors should be collected in a single meta's object. When running the parser with cache this meta's object is read from the cache. Allowing the parser to process just the files that are not changed. And still, be able to link to all files from the newly parsed files. So basically it works like some kind of table-of-contents for the full documentation. 

I think you are right that there are some situations where this logic breaks. I'm not sure in what situations this is as it was already implemented in some way in the original project. And we tried to keep it as is. I think this happens when a fragment or include is processed. Although I think we can fix this by moving the meta's into the parser context. As the context should be passed around when processing a single file including all content imported into that file. 

I do like your suggestion to move the meta's into the transformation phase. This makes a lot of sense, as it would allow us to remove some more logic from the parser. Making it a more pure parser.

wouterj:
Hi! No worries about the response time. OSS time comes and goes :) I think we both understand the value of this freedom, so don't feel obligated to respond in time (whatever "in time" means).

---

> In theory you are doing the same thing, as you added the transformers just before the parser returns the document. If we would move this out of the parser completely. We could introduce a compiling step allowing the transformers to iterate over the full documentation set?

At first, I did like your suggestion of moving this outside the parser. However, I'm now doubting it a bit.
I think it comes down to: what is the deliverable of the parser? Do we want it to return an incomplete/messy AST that is closest to the plain text, or do we want it to return a complete AST?

To give an example of what I mean with an incomplete AST, let's take a look at one of the more complex directives in rst: `.. class::` It can be used in two ways:

```rst
.. with content
.. class:: important

    Some text

.. without content
.. class:: important

Some text

.. the latter can be used in complex scenarios, e.g. adding a class to an image
.. class:: thumbnail
.. image:: images/foo.png
```

The first two examples are rendered equally (`<p class="important">Some text</p>`) and the last example is ``).

The docutils parser fixes this in quite a nice way by producing a `PendingNode` if the class directive has no content. A transformer then visits all pending nodes, adds the class to the adjacent node and removes the pending node from the tree.

If we move the transformer step from the Parser, the Parser will return an AST with "pending nodes". This means the AST is not ready to be rendered, i.e. "incomplete"/messy. I think I favor a parser that returns a complete AST, where these cases are resolved already.

---

> We could introduce a compiling step allowing the transformers to iterate over the full documentation set? So you could even extract information from all documents and manipulate the documents together?

If I continue my reasoning above, I would say this is a different use-case and it makes sense to both have a "transformer" step inside the parser for each document, and a "compiler" step that does things on the project as a whole (and not individual documents).

Btw, we can't use Sphinx as a reference for e.g. the table of contents feature. In Sphinx, document discovery and TOC go hand in hand: You only tell Sphinx about one index file, e.g. `index.rst`. This file then contains a `.. toctree::` directive, listing sub files for this document. Sphinx then parses these files (which in turn can also contain a `.. toctree::` directives). In the end, this automatically creates a full TOC tree like:

```
index.rst
├─ security.rst
│  ├─ security/custom_authenticators.rst
│  ├─ security/voters.rst
│  └─ security/...
├─ controllers.rst
│  └─ controllers/...
├─ ...
```

---

When I find some time for this project again, I'll try to dig a bit deeper in the Metas object and see if it's API can be made more simple and support both global and local reference scopes.

jaapio:
Before anyone continues with this PR, https://github.com/phpDocumentor/guides/pull/132 is improving the metas, I'm still working on this and it's far from finished. But it's getting shape.

jaapio:
With the new meta's structure in place, we might need to revisit this PR; maybe we should extend the Metas class to act like a repository that allows you to find a document by title or even a sub-section by title. As metas is now a Tree of Entries I think it should be flexible enough to cover all cases.

jaapio:
This PR has been split, into multiple pr's. #132  and #233

PR #32, opened 2022-05-17
WIP: Feature/rst table parsing

jaapio:


wouterj:
I've marked (almost?) all table-related functional tests as incomplete in #140 because of buggy behavior. Would be good to remove the `SKIP ...` line from functional tests are make sure they are green again in this PR.

PR #132, opened 2022-12-13
Feature/document compiler

jaapio:
The compiler is meant to be a post-processor when the parser is finished,
we could see this as part of the parsing part, depending on what you expect from
a parser.

From my point of view, parsing is the first step of a traditional compiler that
has multiple iterations to be able to get into the final result, in our case, an ast
ready to be transformed by the renderer into html.

The parser nodes are cachable, and the compiler will transform them into a new structure that
cannot be cached, as it has inter document links that should be updated once a document is
changed. For example, the titles in a toc-tree.


wouterj:
> The parser nodes are cachable, the compilter will transform them into a new structure that
> cannot be cached, as it has inter document linkes that should be updated once a document is
> changed. For example the titles in a toc-tree.

Good point! This distinction makes sense to me :)

jaapio:
This change also impacts the way metas work. Like the rest of the AST, they are now a tree of nodes, in the case of metas a MetaEntry. This is required because toctree is a tree, and also sections are nested. I figured out that the toctree's in sections are also seen as sub-trees of the section. As the previous implementation didn't support this, we were using the meta's everywhere to reconstruct trees, which made the code way too complex. 

I will need more time to finish this, but the first results look promising.

PR #139, opened 2022-12-24
Metas: Implement section target and conflict resolution

wouterj:
This builds on top of the Metas refactor in #132 

Each section in a document should have its own "hyperlink target", which is used to focus the specific section ones the user clicks on the link in the TOC.

Section hyperlink targets are either implicit (based on the title) or explicit (set by using an anchor node). This PR implements a new compiler pass to resolve conflicting implicit references (which is currently missing in the Doctrine parser, ref doctrine/rst-parser#175).

### DIFF
--- a/packages/guides/src/Compiler/NodeTransformers/TocNodeTransformer.php
+++ b/packages/guides/src/Compiler/NodeTransformers/TocNodeTransformer.php
@@ -64,7 +64,7 @@ final class TocNodeTransformer implements NodeTransformer
         //If Toctree is defined at level 2, and max is 3, only titles of the documents are added to the toctree.
 
         foreach ($document->getChildren() as $child) {
-            yield from $this->buildLevel($child, $document, $depth, $node);
+            yield from $this->buildLevel($child, $document, $depth, $node, true);
         }
     }
 
@@ -80,16 +80,16 @@ final class TocNodeTransformer implements NodeTransformer
         }
 
         foreach ($entry->getChildren() as $child) {
-            yield from $this->buildLevel($child, $document, $depth, $node);
+            yield from $this->buildLevel($child, $document, $depth, $node, false);
         }
     }
 
     /** @return Traversable<Entry> */
-    private function buildLevel(MetaEntry $child, DocumentEntry $document, int $depth, TocNode $node): Traversable
+    private function buildLevel(MetaEntry $child, DocumentEntry $document, int $depth, TocNode $node, bool $isDocumentRoot): Traversable
     {
         if ($child instanceof SectionEntry) {
             yield new Entry(
-                $document->getFile(),
+                $document->getFile().($isDocumentRoot ? '' : '#'.$child->getId()),
                 $child->getTitle(),
                 iterator_to_array($this->buildFromSection($document, $child, ++$depth, $node), false)
             );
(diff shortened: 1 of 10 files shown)


packages/guides/src/Compiler/Passes/ImplicitHyperlinkTargetPass.php
```diff
@@ -0,0 +1,82 @@
...
+
+                    $section->getTitle()->setId($node->getValueString());
+
+                    $document = $document->removeNode($key);
+
+                    continue;
+                }
+
+                if ($node instanceof SectionNode) {
+                    $realId = $sectionId = $node->getTitle()->getId();
```
jaapio, line 52: Should this not be part of the parser? Titles are passed around as reference to sections. Changing them would result into possible bugs? As you do this per document, it feels like it could be done with some parser context.

packages/guides/src/Compiler/NodeTransformers/TocNodeTransformer.php
```diff
@@ -80,16 +80,16 @@ private function buildFromSection(
...
     }
 
     /** @return Traversable<Entry> */
-    private function buildLevel(MetaEntry $child, DocumentEntry $document, int $depth, TocNode $node): Traversable
+    private function buildLevel(MetaEntry $child, DocumentEntry $document, int $depth, TocNode $node, bool $isDocumentRoot): Traversable
     {
         if ($child instanceof SectionEntry) {
             yield new Entry(
-                $document->getFile(),
+                $document->getFile().($isDocumentRoot ? '' : '#'.$child->getId()),
```
jaapio, line 92: I think the anchor should be a new property putting all in the same property makes the url building more complex. As the output format is not known by the compiler. The file url is just a reference to find the correct file and render a link.

jaapio (commented):
Thanks a lot. This pr confirms the need of the compiler I introduced.

Some comments, but I'm open to discuss them

packages/guides/src/Compiler/Passes/ImplicitHyperlinkTargetPass.php
wouterj: Ah, I wasn't aware that we used titles in that way currently.

I would say that we should use the new Metas as the only source-of-truth for references. I think we should somehow pass around a "section id" for references and resolve it using the Metas to a real link in a node transformer (?).

This is hard (and probably not wanted) to implement in a parser, as it would require knowing all explicit references and titles in the document. See e.g. these 2 examples:

```rst
.. hyperlink target: some-title

Some Title
----------

...

.. hyperlink target: some-title-1 (resolve conflict with already known titles)

Some Title
----------
```

```rst
.. hyperlink target: some-title-1 (resolve conflict with explicit hyperlink target below)

Some Title
----------

...

.. _some-title:

Example text.
```

packages/guides/src/Compiler/NodeTransformers/TocNodeTransformer.php
wouterj: Agreed, I'm not used yet to think about different output formats :)

packages/guides/src/Compiler/Passes/ImplicitHyperlinkTargetPass.php
jaapio: You are updating the title id, I guess we need to have IDs equal everywhere. So maybe it's not an issue? 

If I get it right, the order matters, so links will break anyway when a new anchor/title is added

PR #140, opened 2022-12-24
Cleanup functional tests

wouterj:
The current CI is failing because all functional tests of doctrine/rst-parser are copy-pasted to this repository. It is actually a nice way to figure out how far this library is with feature parity.

To keep this as a short of todo list, but without making CI hard to use, I propose to mark all failing cases as incomplete (I started with some). This makes sure we don't introduce new unexpected behavior in upcoming PRs (for the working cases).

Meanwhile, I also already found some bugs. Most of them relate to whitespace handling (see also https://github.com/symfony-tools/docs-builder/pull/80#issuecomment-805878833).

### DIFF
--- a/tests/tests/definition-list/definition-list.html
+++ b/tests/tests/definition-list/definition-list.html
@@ -1,27 +1,33 @@
+SKIP bug in complex definition list cases
 <div class="section" id="test">
     <h1>
         Test
     </h1>
-    <div class="note">
-        <p>Test</p>
+    <div class="phpdocumentor-admonition -note">
+        <svg class="phpdocumentor-admonition__icon" fill="none" stroke="currentColor" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+            <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M7 8h10M7 12h4m1 8l-4-4H5a2 2 0 01-2-2V6a2 2 0 012-2h14a2 2 0 012 2v8a2 2 0 01-2 2h-3l-4 4z"></path>
+        </svg>
+        <article>
+            <p>Test</p>
+        </article>
     </div>
     <p>Text around the definition list.</p>
     <dl>
         <dt>term 1</dt>
-        <dd> Definition 1 </dd>
+        <dd>Definition 1</dd>
         <dt>term 2</dt>
         <dd>
             <p class="first">Definition 1</p>
             <p>Definition 2</p>
             <p class="last">Definition 3</p>
         </dd>
-        <dt> term 3 <span class="classifier-delimiter">:</span> <span class="classifier">classifier</span> </dt>
-        <dd> Definition 1 </dd>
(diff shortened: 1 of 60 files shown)


jaapio:
Nice, thanks for this effort. It makes sense to get the build green.

tests/tests/definition-list/definition-list.html
```diff
@@ -1,27 +1,33 @@
+SKIP bug in complex definition list cases
 <div class="section" id="test">
     <h1>
         Test
     </h1>
-    <div class="note">
-        <p>Test</p>
+    <div class="phpdocumentor-admonition -note">
```
wouterj, line 6: The `-note` CSS class looks a bit weird to me, but it is what the templates currently produce. This might need to be fixed later.

wouterj:
Functional tests are green/yellow now!

tests/tests/standalone-email-addresses/standalone-email-addresses.html
```diff
@@ -1,3 +1,4 @@
+SKIP
```
wouterj, line 1: This one completely messes up rSt syntax by using a custom plain-text format as if it was rSt.

Also, it might be taken the "automatic email" detection a bit too far by requiring all RFC 5322 addresses to be correctly detected?

jaapio (approved):
This pr will help us to focus on the correct issues. Thanks for making things more clearly.

Good to be merged :+1:

tests/tests/definition-list/definition-list.html
jaapio: In some future PR, we might want to remove the phpDocumentor-specific template stuff from this repo, in order to deliver a clean template.

tests/tests/standalone-email-addresses/standalone-email-addresses.html
jaapio: Do you think we should remove this from the span parser? Or is it just the regex that's a bit too greedy? 

Please note that the span handling is on my list to refactor because it's still messy. Spans should be a list of inline nodes. This will allow us to do replacements with the new compiler, making the whole span thing a bit less magic. :-)

wouterj: Well, I'm not sure what the official spec says about which e-mailaddresses or websites are automatically detected, but I would suggest to only have a very simple regex for the 80% use-case.

In the very complex 20% cases, people can either wrap it in `<...>` to be detected as hyperlink or use `` `some@email.com <mailto:...>`_ `` for e-mails.

wouterj: For your information, my todo list for this repo at the moment looks like:

1. Finish #139 
2. Finish #16 
3. Fix definition list and `code-block-diff` functional tests (as I've contributed both features to the rst-parser, it's probably the most efficient if I would port those changes over to this repo)

PR #155, opened 2023-01-06
Refactor and improve definition list

jaapio:
The parser of the definition list was quite limited, it didn't support all feature that a definition list should support, and didn't fit in the new node structure.

By changing the implementation, upgrading the test result to cover the sphinx output this new rule is much closer to what we expect according to the doc-utils spec and sphinx output.


wouterj:
> According to the spec, a paragraph is surrounded by blank lines. And a definition ends with a blank line. There is no way to have multiple paragraphs in a single definition. Unless you include a directive or any other block notation.

The spec explicitly allows it though ([ref](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists)):

> A definition is a block indented relative to the term, and *may contain multiple paragraphs and other body elements*. [...]
>
>     [...]
>
>     term 2
>         Definition 2, paragraph 1.
>
>         Definition 2, paragraph 2.

I believe the Sphinx parser looks at the line after the blank line (IIRC, this is also how I implemented it in the Doctrine rst-parser). If the line following the blank line is indented, it is part of the definition. If it isn't, it is terminating the definition list.

tests/tests/definition-list/definition-list.html
```diff
@@ -16,22 +15,18 @@ <h1>
         <dt>term 1</dt>
         <dd>Definition 1</dd>
         <dt>term 2</dt>
-        <dd>
```
wouterj, line 19: I believe the first 3 cases *are* according to spec (or at least, in sync with what Sphinx is producing).

jaapio:
I must have misinterpreted some parts of the spec. It's not possible this way to have multiple definitions with this same term. I think that's where I did go wrong. The HTML spec allows that.  But it doesn't have to be the same for rst. 

The node structure allows it now. I think I will keep it that way. But the rule needs an adjustment. That should be easy, as I implemented the definition parsing as a separate process.

PR #156, opened 2023-01-08
Fix issue with multi paragraph definitions

jaapio:
Definitions can have multiple paragraphs.


jaapio:
@wouterj This should fix the issue introduced in #155 . Multiple paragraphs are now allowed.

PR #213, opened 2023-03-03
Migrate the Intersphinx Inventor Loader t3docs/intersphinx

linawolf:


packages/guides/tests/unit/Intersphinx/InventoryLoaderTest.php
```diff
@@ -0,0 +1,69 @@
...
+use function file_get_contents;
+use function json_decode;
+use function PHPUnit\Framework\assertIsString;
+
+use const JSON_THROW_ON_ERROR;
+
+final class InventoryLoaderTest extends TestCase
+{
+    private InventoryLoader $inventoryLoader;
+    /** @var JsonLoader|MockObject */
```
jaapio, line 20:
```suggestion
    /** @var JsonLoader&MockObject */
```
This will help phpstan to understand it better. Your object is both a `JsonLoader` and `MockObject`

packages/guides/src/Intersphinx/JsonLoader.php
```diff
@@ -0,0 +1,44 @@
...
+use function json_decode;
+
+use const JSON_THROW_ON_ERROR;
+
+class JsonLoader
+{
+    /** @return array<mixed> */
+    public function loadJsonFromUrl(string $url): array
+    {
+        $jsonString = file_get_contents($url);
```
jaapio, line 21: Please keep in mind that `file_get_contents` will only accept URLs when ` allow_url_fopen` in the PHP.ini is enabled. This is the default, but we might want to add an alternative way to do this.

Would the usage of an HTTP client interface be an option? https://www.php-fig.org/psr/psr-18/

We could add a basic implementation of this interface, just doing a `file_get_contents` as a sane default. By using HTTP clients we could even allow people to link to documentationsets with basic authentication, or more advanced authentication if they implement their own way. 

All these are just suggestions, no need to include that right now.

packages/guides/src/Intersphinx/JsonWriter.php
```diff
@@ -0,0 +1,21 @@
...
+use const JSON_PRETTY_PRINT;
+use const JSON_THROW_ON_ERROR;
+
+class JsonWriter
+{
+    /** @param array<mixed> $json */
+    public function saveJsonToFile(array $json, string $filename): void
+    {
+        $jsonString = json_encode($json, JSON_PRETTY_PRINT, JSON_THROW_ON_ERROR);
+        file_put_contents($filename, $jsonString);
```
jaapio, line 19: It would be nice if you could use a `\League\Flysystem\FilesystemInterface` here to write to disk. In phpDocumentor, we are abstracting the filesystem completely because we want to provide options to write to s3 in the future. 

https://flysystem.thephpleague.com/v1/docs/usage/filesystem-api/

If you turn this into a new `\phpDocumentor\Guides\Renderer` implementation it is covered automatically for you.

packages/guides/src/Intersphinx/InventoryLink.php
```diff
@@ -0,0 +1,52 @@
...
+    private string $path;
+    private string $title;
+
+    public function __construct(
+        string $project,
+        string $version,
+        string $path,
+        string $title
+    ) {
+        $this->project = htmlspecialchars($project);
```
jaapio, line 23: I think you don't need the `htmlspecialchars` because escaping is done by twig by default. Or is there another reason to do this?

composer.json
```diff
@@ -30,6 +30,7 @@
     "require": {
         "php": "^7.4||^8.0",
         "ext-mbstring": "*",
+        "ext-json": "*",
```
jaapio, line 33: please use `composer require ext-json:*` to update the lockfile. 

and do also add this requirement to `packages/guides/composer.json` as that is the composer.json used on packagist.

jaapio (commented):
It's nice to see progress, thanks! 

I added some comments to help you get an understanding of what we are doing. Feel free to reach out, and discuss the comments. None of them are strict requirements, but mostly things to keep in mind.

packages/guides/src/Intersphinx/JsonWriter.php
linawolf: Actually the json writer is not really needed as of yet, so I would remove it and turn to the renderers later on

packages/guides/src/Intersphinx/InventoryLink.php
linawolf: Ok, lets delegate this to the output then

linawolf:
@jaapio I don't really get what the ComposerRequireChecker  is doing here and why it is red?

jaapio:
Composer-require-checker checks for dependencies used by our code but not part of the composer.json. 

So for example when you are using http client interface which is a dependency of symfony http client. It will fail if you do not require http client as a package. 

This will help you to require all direct dependencies of a project.

jaapio (approved):
This is good to go, we can do other parts in a separate pr.

Issue #215, opened 2023-03-04
Drop PHP 7.2 and 7.3 Support from guides?


linawolf:
In the main composer.json we are requiring php above 7.4, and tests are run for PHP 7.4 and 8.0 as far as I can tell.
packages/guides/composer.json however still allows PHP 7.2. I would really like to use typed properties and as this version is also not covered by tests anymore can we drop it?

jaapio:
Yes we should drop those versions. 

I would like to keep php 7.4 support for now, but we can drop that in a year

wouterj:
I'm not sure about phpDocumentor's policies, but given this is project has the experimental status and PHP 7.4 is eom for 3 months already, I would suggest even bumping the minimum to 8.0 (if the works for the projects working on it of course).

jaapio:
We see a lot of people using phpDocumentor on eol versions. So we keep support a bit longer. 

If it becomes a hassle to keep support old php versions we can drop them. But for now, 7.4 it targeted.

PR #216, opened 2023-03-04
Drop PHP 7.2 and 7.3 Support from guides

linawolf:
resolves https://github.com/phpDocumentor/guides/issues/215


jaapio:
Could you please update https://github.com/phpDocumentor/guides/blob/main/packages/guides-restructured-text/composer.json#L20 as well?

linawolf:
@jaapio done

PR #218, opened 2023-03-05
Add functional tests from doctrine/rst-parser

linawolf:
that have been added since the fork

PR #220, opened 2023-03-05
Add a complex container test

linawolf:
Including a list in the content

A simple container test is already provided by #218


jaapio:
Good to see that the more complex version also works!

PR #221, opened 2023-03-05
Make class directive work for simple cases

linawolf:
class directives applied to neighboring nodes post-phoned for now


wouterj:
> class directives applied to neighboring nodes post-phoned for now

When the changes from #21 are merged (I believe Jaap started working based on my commits there), this should be relatively easy by looping over the node tree and applying class nodes to the next node.

jaapio:
The compiler is already there. So this can be done with a new compiler pass.

jaapio:
I would suggest adding a none rendered node that represents the class in the node tree. This will help us to do some compiling after the parser is ready. 

Let's suggest that we have a `ClassNode` that contains the class / classes set by the directive. 

Then we could implement a new `\phpDocumentor\Guides\Compiler\NodeTransformer` like this. 

```
final class ClassNodeTransformer implements \phpDocumentor\Guides\Compiler\NodeTransformer
{
    public function enterNode(Node $node): Node
    {
       if ($node instanceOf ClassNode) {
          $this->class = $node->getClasses();
          return $node;
       }
       
       $node->setClasses($this->class);
    }

    public function leaveNode(Node $node): ?Node
    {
        if ($node instanceOf ClassNode) {
           //Remove the class node from the tree. 
           return null;
        }
    }

    public function supports(Node $node): bool
    {
        //Should we do something here to detect nodes are supported? Or could every node have a class?
        return true;
    }
} 
```

I did not test this and the code may contain issues, but I think it could be something like this. 

If you need more complex logic I would recommend you have a look at the `\phpDocumentor\Guides\Compiler\CompilerPass` interface, a compiler pass will get all Documents and allows you to apply more complex transformations. But if you think a node transformer might work please use that. As the compiler passes are more complex that will leave you on your own about our AST.

packages/guides/src/Compiler/NodeTransformers/ClassNodeTransformer.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @var string[]
+     */
+    private array $classes = [];
+
+    public function enterNode(Node $node): Node
+    {
+        if ($node instanceof ClassNode) {
+            $this->classes = $node->getClasses();
```
jaapio, line 23: We do reset classes here, but should they not be reset when you start a new document?

jaapio:
Appart from the question about the reset of classes, this is great! thanks for this contribution

packages/guides/src/Compiler/NodeTransformers/ClassNodeTransformer.php
linawolf: I tackled the issue of resetting and added tests for those scenarios

jaapio:
Cool to have this feature in. Thanks!

PR #222, opened 2023-03-05
Nest overlined or framed titles different from underlined ones

linawolf:
TYPO3 docs usually have the Document title between two lines of `=` each


jaapio:
Nice, thanks!

PR #223, opened 2023-03-05
Improve comment and title / separator handling

linawolf:
See https://docutils.sourceforge.io/docs/user/rst/quickref.html#comments

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TitleRule.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TitleRule.php
@@ -17,11 +17,12 @@ use phpDocumentor\Guides\Nodes\CompoundNode;
 use phpDocumentor\Guides\Nodes\Node;
 use phpDocumentor\Guides\Nodes\TitleNode;
 use phpDocumentor\Guides\RestructuredText\Parser\DocumentParserContext;
+use phpDocumentor\Guides\RestructuredText\Parser\LineChecker;
+use phpDocumentor\Guides\RestructuredText\Parser\LinesIterator;
 use phpDocumentor\Guides\RestructuredText\Span\SpanParser;
 
 use Symfony\Component\String\Slugger\AsciiSlugger;
 use function in_array;
-use function strlen;
 use function trim;
 
 /**
@@ -30,41 +31,7 @@ use function trim;
  */
 class TitleRule implements Rule
 {
-    private const HEADER_LETTERS = [
-        '!',
-        '"',
-        '#',
-        '$',
-        '%',
-        '&',
-        '\'',
-        '(',
-        ')',
-        '*',
-        '+',
-        ',',
-        '-',
-        '.',
-        '/',
-        ':',
-        ';',
-        '<',
-        '=',
-        '>',
-        '?',
-        '@',
-        '[',
-        '\\',
-        ']',
-        '^',
-        '_',
-        '`',
-        '{',
-        '|',
-        '}',
-        '~',
-    ];
-
+    private const TITLE_LENGTH_MIN = 2;
     private SpanParser $spanParser;
 
     public function __construct(SpanParser $spanParser)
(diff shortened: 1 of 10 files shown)


packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TransitionRule.php
```diff
@@ -110,7 +110,8 @@ public function isSpecialLine(string $line): ?string
 
     private function currentLineIsASeparator(string $line, ?string $nextLine): ?string
     {
-        $letter = $this->isSpecialLine($line);
+        // Separators must have a minimum length of 3, so they don't apply to `::` (code block) or `..` comment
+        $letter = $this->isSpecialLine($line, 3);
```
jaapio, line 114: According to the spec, this should actually by 4 chars.

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TransitionRule.php
```diff
@@ -87,9 +87,9 @@ public function apply(DocumentParserContext $documentParserContext, ?CompoundNod
         return new SeparatorNode(1);
     }
 
-    public function isSpecialLine(string $line): ?string
+    public function isSpecialLine(string $line, int $minimumLength = 2): ?string
```
jaapio, line 90: I think this method is never used by any other class. If not, please make it private :-D

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TitleRule.php
```diff
@@ -142,7 +142,8 @@ public function isSpecialLine(string $line): ?string
     private function currentLineIsAnOverline(string $line, ?string $nextLine): string
     {
         $letter = $this->isSpecialLine($line);
-        if ($nextLine !== null && $letter && $this->isTextLine($nextLine)) {
+        if ($nextLine !== null && $letter && $this->isTextLine($nextLine) &&
```
jaapio, line 145: We can replace `$nextLine !== null && $letter && $this->isTextLine($nextLine)` with `LineIterator::isNullOrEmptyLine($nextLine) === false`

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TitleRule.php
```diff
@@ -142,7 +142,8 @@ public function isSpecialLine(string $line): ?string
...
+            (strlen($line) > 2 || strlen($line) >= strlen($nextLine))) {
```
jaapio, line 146: Please use `mb_strlen` as the title might contain multibyte chars. 

Should the `strlen($line) > 2` not be part of the `isSpecialLine` method? this will make this statement more readable.

jaapio (changes requested):
Some minor improvements to make the code a bit more readable.

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TitleRule.php
linawolf: No not really. Letter is working on $line while the other two are refering to $nextline

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TransitionRule.php
linawolf: It is duplicated in Transition and title, however, so I extracted it

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/TitleRule.php
linawolf: Refactored the code to make it easier to understand

jaapio (approved):

jaapio:
Nice, thank you. Yet another step closer to a full RST parser

PR #225, opened 2023-03-05
Directive ends on complete unindent

linawolf:
Expecially in code blocks the first line of the directive content can have more indentation then the following ones. Overall this is true for all directives.

The general rule is that a directive ends when the indentation level of the directive declaration is found or when the file ends.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/Buffer.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/Buffer.php
@@ -84,4 +84,13 @@ class Buffer
     {
         $this->lines = [];
     }
+
+    public function unIndent(int $indentation): void
+    {
+        array_walk($this->lines, function (&$value) use ($indentation) {
+            if (strlen($value) >= $indentation) {
+                $value = substr($value, $indentation);
+            }
+        });
+    }
 }
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveRule.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveRule.php
@@ -97,7 +97,7 @@ final class DirectiveRule implements Rule
         // Processing the Directive, the handler is responsible for adding the right Nodes to the document.
         try {
             return $directiveHandler->process(
-                $documentParserContext->withContents($buffer->getLinesString()),
+                $documentParserContext->withContentsPreserveSpace($buffer->getLinesString()),
                 $directive->getVariable(),
                 $directive->getData(),
                 $directive->getOptions()
@@ -215,20 +215,17 @@ final class DirectiveRule implements Rule
     private function collectDirectiveContents(LinesIterator $documentIterator): Buffer
     {
         $buffer = new Buffer();
-        $indenting = 1;
-        while (LinesIterator::isBlockLine($documentIterator->getNextLine(), $indenting)) {
+        $minIndenting = PHP_INT_MAX;
+        while (LinesIterator::isBlockLine($documentIterator->getNextLine())) {
             $documentIterator->next();
             $line = $documentIterator->current();
-            if ($indenting === 1 && LinesIterator::isEmptyLine($line) === false) {
+            if (LinesIterator::isEmptyLine($line) === false) {
                 $indenting = mb_strlen($line) - mb_strlen(ltrim($line));
+                $minIndenting = min($minIndenting, $indenting);
             }
-
-            if ($line !== '') {
-                $line = substr($documentIterator->current(), $indenting);
-            }
-
             $buffer->push($line);
         }
+        $buffer->unIndent($minIndenting);
         return $buffer;
     }
 }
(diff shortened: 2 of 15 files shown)


packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveRule.php
```diff
@@ -184,20 +184,19 @@ private function parseDirectiveOption(string $line): DirectiveOption
     private function collectDirectiveContents(LinesIterator $documentIterator): Buffer
     {
         $buffer = new Buffer();
-        $indenting = 1;
-        while (LinesIterator::isBlockLine($documentIterator->getNextLine(), $indenting)) {
+        $minIndenting = 1000;
```
jaapio, line 187: Could you please extract the default value to a constant, so we can add docblocks above it to explain what it is? Right now it feels like a random number.

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveRule.php
```diff
@@ -184,20 +184,19 @@ private function parseDirectiveOption(string $line): DirectiveOption
...
+                $minIndenting = min($minIndenting, $indenting);
             }
-
-            if ($line !== '') {
-                $line = substr($documentIterator->current(), $indenting);
-            }
-
             $buffer->push($line);
         }
+        if ($minIndenting < 1000) {
```
jaapio, line 197: Is this not equal to always?

packages/guides-restructured-text/src/RestructuredText/Parser/Buffer.php
```diff
@@ -84,4 +84,13 @@ public function clear(): void
     {
         $this->lines = [];
     }
+
+    public function unIndent(int $indentation): void
```
jaapio, line 88: I get what you are doing here, but it feels like something that can be forgotten when using this class. Should we add this in the `getLinesString` method and `getLines`? with the default of indenting 1?

jaapio (changes requested):
A few suggestions to improve the code. 

And it would be nice if we had more unit tests. The functional tests are beneficial, but they are much slower. So if possible it would be nice to add some tests that cover the edge cases of the `DirectiveRule` You can have a look at the other Rule tests for inspiration.

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveRule.php
linawolf: It is a random number it just needs to be very high...

packages/guides-restructured-text/src/RestructuredText/Parser/Buffer.php
linawolf: Mmh, I don't really get what you suggest..

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveRule.php
linawolf: Mmh, maybe with PHP_INT_MAX ?

packages/guides-restructured-text/src/RestructuredText/Parser/Buffer.php
jaapio: I would like to see the effect of this method, be part of the `getLinesString` method and `getLines`

eg. 

```
getLines(int $indentation = 1)
{
   return array_map(function($line) {
            if (strlen($line) >= $indentation) {
                return substr($line, $indentation);
            }
            
            return $line;
      }
   }
}
```
This way we can not forget to remove the indentation, we have multiple places where we use the buffer and strip the first part of the indenting before we start another cycle of parsing the content of a block. Those can be replaced this way.

linawolf: In the very place where I used the unIndent function getLines is not used, just the buffer returned. So we would have to return the information about the depth of the indentation and the buffer only to apply unindent lateron?

jaapio: I have created a list of possible locations to dig into: 

- This method removes the indenting like you do. Based on the first line of a block.
 `\phpDocumentor\Guides\RestructuredText\Parser\Productions\BlockQuoteRule::normalizeLines`
- Here I do something simular for lists
  `packages/guides-restructured-text/src/RestructuredText/Parser/Productions/ListRule.php:107`
- `packages/guides-restructured-text/src/RestructuredText/Parser/Productions/EnumeratedListRule.php:109`
- Definition lists: `packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DefinitionListRule.php:99`

jaapio: Yes, that would work for me. I will turn this into an issue.

jaapio (approved):

PR #230, opened 2023-03-07
Enable multiline directive options

linawolf:
directive options are basically field lists:

https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives 'Directive options are indicated using field lists; the field names and contents are directive-specific.'

Field list options can have multiple lines:
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists


jaapio:
Thanks this is great!

PR #233, opened 2023-03-07
Ref role resolving

jaapio:
The text role `ref` is now supported as it should.
Simplyfied the way links are rendered without markup in the titles.
This allows us again to create links using just text.

There is a major issue with using the subparsers in this change
the sub documents are isolated from the main document. This causes
issues handling includes. We should refactor this in another effort.


jaapio:
@wouterj  this supersedes the PR you created #21 and should fixes #16

jaapio:
@linawolf I would like to have your opinion on the subparser. I think we do not need it anymore as we have a different way of doing this using the `\phpDocumentor\Guides\RestructuredText\Parser\Productions\RuleContainer` you can see an example of this in `\phpDocumentor\Guides\RestructuredText\Parser\Productions\SimpleTableRule`.

Right now `Admonitions` might contain everything, but I think in the should not have all nodes and directives. Some directives and nodes are not allowed to be used in directives. following https://docutils.sourceforge.io/docs/ref/doctree.html there are categories of elements that I tried to model using the parser setup. 

And for example, the includes do not work as they should, as we should literally include the nodes of the included file. So title levels are kept the same when a document is included.

wouterj:
Thanks for continuing my work! I'll likely try these changes out in the Symfony docs and do a review at Thursday

linawolf:
I am honestly still getting used to the new architecture by fixing some tiny things. Not far enough in to have an oppinion, really

wouterj:
Not sure if it's 100% related to this one, but when rendering the Symfony docs on this branch (rebased on #252) there are some invalidly detected references.

Examples:

``` ``choice``: renders one, two (default) or three select inputs (hour, minute, ```
everything after `:` is detected as a single reference (there is no reference on the line).

``` When using different values for `model_timezone`_ and ``view_timezone``, ```
this whole line is detected as a single reference (instead of only `model_timezone`)

``` **type**: ``integer`` **default**: Locale-specific (usually around ``3``) ```
` Locale-specific (usually around ``3``) ` is detected as reference

``` Securing by an Expression ```
gives an error "invalid cross reference: ecuring by an Expression"

jaapio:
I think this are bugs in the span parser. I need to have a look. But not related to this pr.

PR #241, opened 2023-03-10
Setup a simple application to run guides

jaapio:
This is a simple application allowing people to
run guides as a standalone application.


wouterj:
Interesting, I've been maintaining something similar (as a start of a symfony-tools/docs-builder port that uses this library) to better test out guides for the Symfony docs: https://github.com/wouterj/symfony-docs-guides/tree/main/lib/docs-builder

jaapio:
Oh nice, I was looking how to solve some issues, your project could be a source. 

Thanks for your response

Issue #219, opened 2023-03-05
Add doctrine-rst-parser functional build tests


linawolf:
In the doctrine/rst-parser we introduced a new category of functional tests that work on a complete document tree:

https://github.com/doctrine/rst-parser/tree/0.6.x/tests/Functional/tests/build

This kind of test is not supported in guides yet as far as I can tell

jaapio:
It would be good to have some tests that are testing more than isolated nodes. So I would welcome a PR for this.

PR #243, opened 2023-03-11
Introduce basic integration tests

linawolf:
references #219

### DIFF
--- a/tests/Integration/IntegrationTest.php
+++ b/tests/Integration/IntegrationTest.php
@@ -1,0 +1,179 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\RST\Integration;
+
+use League\Flysystem\Adapter\Local;
+use League\Flysystem\Filesystem;
+use phpDocumentor\Guides\Compiler\Compiler;
+use phpDocumentor\Guides\Compiler\DocumentNodeTraverser;
+use phpDocumentor\Guides\Compiler\NodeTransformers\TocNodeTransformer;
+use phpDocumentor\Guides\Compiler\Passes\MetasPass;
+use phpDocumentor\Guides\Compiler\Passes\TransformerPass;
+use phpDocumentor\Guides\FileCollector;
+use phpDocumentor\Guides\Handlers\ParseDirectoryCommand;
+use phpDocumentor\Guides\Handlers\ParseDirectoryHandler;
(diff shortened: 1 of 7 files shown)


tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+        foreach ($documents as $document) {
+            try {
+                $renderDocumentHandler->handle(
+                    new RenderDocumentCommand(
+                        $document,
+                        RenderContext::forDocument(
+                            $document,
+                            $sourceFileSystem,
+                            new Filesystem(new Local($outputPath)),
```
jaapio, line 129: If we would replace this local adapter with a memory adapter, we could get rid of the removal statements at the start of this test. 

https://flysystem.thephpleague.com/v1/docs/adapter/memory/

It will require some adjustments to the asserts, as the files are not on disk anymore, but this will prevent issues when the tests are failing for some reason.

tests/Integration/IntegrationTest.php
```diff
@@ -0,0 +1,178 @@
...
+                            '/',
+                            $metas,
+                            new UrlGenerator(),
+                            'html'
+                        )
+                    )
+                );
+            } catch (\Exception $e) {
+                echo "Error:" . $e->getMessage() . PHP_EOL;
```
jaapio, line 138: Should we catch exceptions here? It would make sense to make tests fail on those.

jaapio (commented):
Some small improvements to make the tests run faster, and more stable. 

It might be good to adjust the gitlab workflow to run these test in CI. 

```
  integration-tests:
    name: "integration test"
    uses: "phpDocumentor/.github/.github/workflows/continues-integration.yml@v0.3.0"
    needs: "unit-tests"
    with:
      php-versions: "['7.4', '8.0', '8.1', '8.2']"
      test-suite: "integration --verbose"
```

tests/Integration/IntegrationTest.php
linawolf: I would like the integration tests to be as close to the real thing as possible and this also means writing files to disk.

Once you moved forward with the standalone application I would like to use the same logic here. I would rather have a cleanup make method deleting all the temp folders.

tests/Integration/IntegrationTest.php
linawolf: You are right.

PR #245, opened 2023-03-11
Add test that image gets copied

linawolf:
depends on https://github.com/phpDocumentor/guides/pull/243


jaapio (approved):

linawolf:
@jaapio please revew again. As @mvriel asked for I added tests with relative and absolute image paths

PR #248, opened 2023-03-12
Inventory Index output for intersphinx

linawolf:


jaapio:
I get the feeling when looking at the changes you are making you are trying to push this at the wrong level. There are 2 things you need.

- metas
- output directory. 

If I try to translate that into what we have right now there is no such match. In this library right now. Like we are missing some level of rendering. And I think I noticed this before on my side as well. We need something that's more overlapping with what the parser does. This is split into 2 levels. 

- Directory parsing
- Document parsing

In this case you want something like `Directory rendering` but without iterating over the documents, but just provide the full set of documents including the metas. to render everything. In phpDocumentor itself, we do have such a thing, but it is quite close to the implementation we have there. But in essence, this could be the thing you are missing. 

In free form translated:

```php
final class RenderCommand
{
    /** @var DocumentNode[] */
    private $documentationSet;

    /** @var FilesystemInterface */
    private $origin;

    /** @var FilesystemInterface */
    private $destination;

    public function __construct(
        array $documentationSet,
        Metas $metas
        FilesystemInterface $origin,
        FilesystemInterface $destination
    ) {
        $this->destination = $destination;
        $this->documentationSet = $documentationSet;
        $this->origin = $origin;
        $this->metas = $metas;
    }

    public function getDocumentationSet(): array
    {
        return $this->documentationSet;
    }

    public function getOrigin(): FilesystemInterface
    {
        return $this->origin;
    }

    public function getDestination(): FilesystemInterface
    {
        return $this->destination;
    }

    public function getDestinationPath(): string
    {
        return $this->documentationSet->getOutputLocation();
    }

    public function getTargetFileFormat(): string
    {
        return $this->documentationSet->getOutputFormat();
    }
}
```
The command here will just specify: Documents, Metas and filesystems. 

The handler could get a set of OutputFormatStategies. 

- html
- tex
- intersphinx

Where each of them will implement a method `supports($command->getTargetFileFormat(): bool`

In case of the HTML and tex strategy it will do the looping that we currently have in the example file:

```php
foreach ($documents as $document) {
    echo "Render: " . $document->getFilePath() . PHP_EOL;

    try {
        $renderDocumentHandler->handle(
            new RenderDocumentCommand(
                $document,
                RenderContext::forDocument(
                    $document,
                    $sourceFileSystem,
                    new Filesystem(new Local(__DIR__ . '/out')),
                    '/example/',
                    $metas,
                    new UrlGenerator(),
                    'html'
                )
            )
        );
    } catch (Exception $e) {
        echo "Error:" . $e->getMessage() . PHP_EOL;
    }
}
```
In your case it would have something like, and fully ignores the documents. 

```php
        $inventory = [
            'std:doc' => []
        ];
        foreach ($environment->getMetas()->getAll() as $key => $documentEntry) {
            $inventory['std:doc'][$key] = [
                '',
                '',
                preg_replace('/^\//', '', $environment->relativeDocUrl($documentEntry->getFile())),
                $documentEntry->getTitle()->toString()
            ];
        }
        return (string) json_encode($inventory, JSON_PRETTY_PRINT);
 ```
        
This approach has 2 benefits.
1. We will make the rendering call much easier, as nobody needs to care about the context creation 
2. We will make your part much easier. as you don't need to care about the context at all. You could just use the `UrlGenerator` to create the URLs.

Tasks todo:
- [ ] Introduce a `RenderCommand`
- [ ] Introduce an Interface for `RenderSet` having 2 methods
      - `supports(string $outputFormat)`
      - `render($documents, $metas, $originFileSystem, $destinationFileSystem)`
- [ ] Introduce a `RenderHandler` handling the `RenderCommand`
       The render handler will find the correct `RenderSet` implementation based on the output format. And call the `render` method

How does this sound to you?

**Comment by linawolf** [comment]
@jaapio the metas do not contain the concrete absolute or relative links to files and titles. We can currently only get them from the context / environment. How should we solve that?

File: packages/guides/src/Renderer/RenderIntersphinxSet.php [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return $outputFormat === RenderSet::INTERSPHINX;
+    }
+
+    public function render(RenderCommand $renderCommand): void
+    {
+        $inventory = [
+            'std:doc' => []
+        ];
+        $renderContext = RenderContext::forSitemap(
```
jaapio, line 25: I think it might be good to spit the RenderContext into 2 separated classes. the only reason you need this here is because of the `relativeDocUrl` method. I'm even thinking if we could extract this function to make it part of the UrlGenerator.

Splitting will give you the benefit of the non-nullable `DocumentNode` in the context.

packages/guides/src/Renderer/RenderSet.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace phpDocumentor\Guides\Renderer;
+
+use phpDocumentor\Guides\Handlers\RenderCommand;
+
+interface RenderSet
```
jaapio, line 7: I'm a bit in doubt about this class name. 

`Renderer` would fit better, but that's too generic. The term `Set` is coming from phpDocumentor. But I'm not sure if it makes sense here. What do you think?

jaapio (commented):
This looks very nice, thanks for this improvement!

jaapio:
If you finalize the implementation I will do the phpstan part.

packages/guides/src/RenderContext.php
```diff
@@ -81,6 +81,28 @@ public static function forDocument(
         return $self;
     }
 
+    public static function forSitemap(
```
jaapio, line 84: This method is not used anymore? So we can remove it?

packages/guides/src/Renderer/TypeRenderer.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace phpDocumentor\Guides\Renderer;
+
+use phpDocumentor\Guides\Handlers\RenderCommand;
+
+interface TypeRenderer
+{
+    public const HTML = 'html';
```
jaapio, line 9: I would like to move these contants to their classes, `HtmlTypeRenderer` and `IntersphinxTypeRenderer` That will make this interface really separated from the actual implementation

packages/guides/tests/unit/Renderer/InventoryRendererTest.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace phpDocumentor\Guides\Renderer;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversDefaultClass \phpDocumentor\Guides\Renderer\InventoryRenderer
+ * @covers ::<private>
+ */
+final class InventoryRendererTest extends TestCase
```
jaapio, line 22: This seems to be not implemented. Please remove this class if you are not planning to add unittests :-)

jaapio (changes requested):
Some minor cleanup work needs to be done. But we are nearly there! 

Thank you so much for this very nice feature!

packages/guides/src/RenderContext.php
linawolf: removed it

packages/guides/tests/unit/Renderer/InventoryRendererTest.php
linawolf: woops. must have been a leftover

jaapio (approved):

PR #252, opened 2023-03-14
Add CompileDocumentsCommand

wouterj:
The command and handler are extremely simple, but makes the usage API of this library a bit more tidy (the service generating the docs now only needs the `CommandBus`, instead of `CommandBus`+`Compiler`).


jaapio:
I like `extremely simple` :-D 

Thanks for this improvement!

PR #254, opened 2023-03-15
Create a basic bootstrap template

linawolf:


mvriel:
I haven't looked at it in detail, but I do have a request: please put as little as possible in the layout.html.twig in terms of content. The reason is that we need to be able to pull the components in into phpDocumentor and be able to provide our own layout around it, including navigation.

You can check the default template of phpDocumentor, a lot of effort has been put into it to make that a component-based structure and allow users to do the same thing and override everything :)

linawolf:
Well you could keep doing that with the standard template couldn't you? There is no need for you to use the bootstrap theme then. However I am more thinking in terms of standalone documentation rendering and I would want a complete site with menu etc then.  I would however split it up more to make it easier extendable like with separate twig includes for header footer, menu etc. Also nothing would hinder you to extend this theme and override the layout.twig.html to just get the plain, bootstrap conform content. The goal is to create an easy to use and extend standard theme and to make more clear of how theeming could work.

mvriel:
The thing is, I don't want to extend the theme: I want to embed the theme :) Guides is just one small part of phpDocumentor, and for phpDocumentor to do its magic it requires its own theme as the basis.

By limiting what you put in the layout and separating it out into components, users are free to use this template both ways. As an extension point or by embedding components

mvriel:
But do stop me if I am shouting to soon by the way :D The template system and templates themselves are my proverbial kids :D I have spent a long time optimizing the templates for reusability and am a bit attached :D

linawolf:
I am not changing your templates. They reside in guides and are the default and they will keep being the default.

tests/Functional/FunctionalTest.php
```diff
@@ -40,6 +41,9 @@ class FunctionalTest extends ApplicationTestCase
 
     protected function setUp(): void
     {
+        self::prepareContainer(new Configuration([
+            __DIR__ . '/../../packages/guides/resources/template/html/guides',
```
jaapio, line 45: Should we not use the default config here, to ensure the default is working as expected. So not passing this array

jaapio (approved):

PR #257, opened 2023-03-20
Introduce the rubric directive

linawolf:
The general structure of directives can - in most cases - be saved into a general node type. 

A directive consists of a type, an argument, one or several options and content. the content can be in the form of nodes or plain text (for example code nodes)

I demonstrate usage of a DirectiveNode by introducing the rubric directive that was still missing. We can possibly replace many of the other nodes for directives with a general directive node.

In Sphinx there are rules for each directive what and how many of each kind of data it accepts. We might want to model that in future. Right now non-expected content is ignored.

The `:class:` option is one that can be added to all directives and it has the same effect like setting the class with the class directive. So I saved the class option in the same variable.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Directives/RubricDirective.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Directives/RubricDirective.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\RestructuredText\Directives;
+
+use phpDocumentor\Guides\Nodes\Node;
+use phpDocumentor\Guides\Nodes\RubricNode;
+use phpDocumentor\Guides\RestructuredText\Parser\Directive;
+use phpDocumentor\Guides\RestructuredText\Parser\DocumentParserContext;
+
+/**
+ * The "rubric" directive inserts a "rubric" node into the document tree. A rubric is like an informal heading
+ * that does not correspond to the document's structure.
+ *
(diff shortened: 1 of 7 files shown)


jaapio:
I like this suggestion, but I would like to point you at the structure of the parser and the existing `GenericNode`. 

**Parser structure**
I could not complete it yet, but the parser is built out of productions. Each production can get a set of productions that will produce the child nodes of a production. This way we can configure the nodes that can be the child nodes of a directive. 
Right now we took a shortcut, by using the original subparser structure, but that should be removed in my opinion.
This is the reason the \phpDocumentor\Guides\RestructuredText\Parser\Productions\DirectiveRule collects all lines in a directive. trims the indenting and passes it to a DirectiveHandler. 

**Nodes**
I do prefer type strict systems and for that reason, I think it makes sense to not reuse the nodes, it's ok to have base classes, and shared configurable renderers. But I think it makes sense to model the nodes as closely as possible to their final type. 

I hope this makes it a bit more clear. I think we are on the same line, but I need to point you at some existing systems to make this work for you. If needed we can have a call to explain it more and discuss the details. meanwhile I will try to make a start with the missing documentation about the structure of the packages.

linawolf:
@jaapio how shall we proceed with this PR?

jaapio:
I think we need to have a chat about this. But I need to prepare some things for that. Because my comment didn't make my statement clear enough. I will come back on this later.

linawolf:
@jaapio, so I just created a node and a directive for the rubric node.

jaapio:
Thanks 👍

PR #258, opened 2023-03-20
Move templates in subdirectories

linawolf:
according to their usage to make them easier to find. Keep all roles together, all directives together et cetera


jaapio (commented):
Suggestion to spit the html and tex files as root folders, to make it easier to add new output types? 

Would it make sense to follow the structure of the nodes in the RST specification? 
https://docutils.sourceforge.io/docs/ref/doctree.html#element-hierarchy

    Structural Elements
        Structural Subelements
            Bibliographic Elements
            Decorative Elements
    Body Elements
        Simple Body Elements
        Compound Body Elements
            Body Subelements
    Inline Elements

I tried to use this same structure in the parser to build the productions. Aligning those structures makes kind of sense in my head. 

Markdown seems to have a comparable structure. Regarding the reparation of inline and body elements. But as it is limited in its syntax it seems to be more flat.

How do you see this?

linawolf:
@jaapio I gave it a trial with the main categories

PR #259, opened 2023-03-21
Make document node rendering more normalized

jaapio:


linawolf (commented):
Looks good to me, my fingers have itched before changing all the $enviroment variables into $renderContext ;)

PR #270, opened 2023-03-25
New span parsing: Setup some example text roles

linawolf:


linawolf:
@jaapio once you are back at your keyboard, is this going in the right direction?

jaapio:
Yes this looks 👍 

We need to introduce other spantoken types for each rule. I guess you already understood that 😁
And when looking at this, i realized that order of the rules might matter.  Because of the overlapping start characters for types of links. 

Thanks for taking this effort

linawolf:
I am at it!

linawolf:
working with negative lookahead / lookbehind to prevent the roles to be misinterpreted. Order does however matter when for example a literal token contains other special chars. A literal token can contain more or less anything but its endstring

jaapio:
We could solve that by adding priorities to the rules and inject them in a priority queue. 

So it doesn't matter in what order they are injected in the InlineMarkupRule. 

Other thing to solve are escaped chars.

linawolf:
In the tokenization we should at least make sure not to split tokens when there are escaped whitespaces. Escape of special chars that can start or end inline markup should be solveable by lookbehind as well

packages/guides-restructured-text/tests/unit/Parser/Productions/InlineMarkup/TextRoleRuleTest.php
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public function ruleAppliesProvider(): array
+    {
+        return [
+            [
+                [':role:`something'],
+                true,
+            ],
+            [
+                [':php:class:`something'],
```
jaapio, line 35: This is a special role. I think sphinx calls this domains: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

The current span parser allows this kind of roles in domains. We are using the `CrossReferenceNode` for this. Not sure how this should work.

jaapio (commented):
In general, this still looks great! I recommend you take the last 2 commits to another branch. As it is not related to this particular change. And the PR is getting big. 

It might be a good idea to start wrapping up and preparing for the merge. We can always improve more. But we have to make small steps to keep the work fun :)

packages/guides-restructured-text/tests/unit/Parser/Productions/InlineMarkup/TextRoleRuleTest.php
linawolf: As we give the oppurtunity to invent any kind of role you can also invent one with a domain. I would not solve this on parser level...

linawolf:
Ok, so let's wrap it up and move the integration part to another PR I also moved the last two commits to #275

PR #285, opened 2023-04-09
Add skipped integration test with a typical TYPO3 Extension documentation

linawolf:


linawolf:
I will provide better tests

Issue #298, opened 2023-04-15
Add docinfo support


jaapio:
https://docutils.sourceforge.io/docs/ref/doctree.html#docinfo

https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists

linawolf:
Currently working on the field-lists

linawolf:
Also support special Sphinx-related field list items:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html

PR #299, opened 2023-04-15
Support field lists

linawolf:
See https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists

resolves https://github.com/phpDocumentor/guides/issues/298


packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldListRule.php
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * @param FieldListItemNode[] $fieldListItemNodes
+     */
+    private function addMetadata(DocumentNode $documentNode, array $fieldListItemNodes): void
+    {
+        foreach ($fieldListItemNodes as $fieldListItemNode) {
+            switch (strtolower(trim($fieldListItemNode->getTerm()))) {
+                // Bibliographic Fields Defined by docutils
+                // https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists
+                case 'abstract':
```
jaapio, line 95: It would be nice to refactor this into a list of factories that we can configure via the constructor. This will allow us to extend this switch without changing this class. 

As sphinx has a different list of supported fields that docutils has. Other standards could have more types?

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/RuleContainer.php
```diff
@@ -49,6 +50,10 @@ public function apply(DocumentParserContext $documentParserContext, CompoundNode
                 $on->addChildNode($newNode);
             }
 
+            if ($production instanceof TitleRule && $on instanceof DocumentNode) {
+                $on->setTitleFound(true);
```
jaapio, line 54: Why do we need this?

packages/guides/src/Nodes/FieldLists/FieldListItemNode.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @extends CompoundNode<Node>
+ */
+final class FieldListItemNode extends CompoundNode
+{
+    private string $term;
+    private string $plaintextContent;
+
+    /**
+     * @param Node[] $children
```
jaapio, line 28: Are all nodes allowed? or just metadata nodes?

packages/guides/src/Nodes/Metadata/OrphanNode.php
```diff
@@ -0,0 +1,25 @@
...
+namespace phpDocumentor\Guides\Nodes\Metadata;
+
+/**
+ * If set, warnings about this file not being included in any toctree will be suppressed.
+ */
+class OrphanNode extends MetadataNode
+{
+    public function __construct()
+    {
+        parent::__construct('true');
```
jaapio, line 23: It feels like we need some boolean metadata node? Or should these be "options" on a document?

packages/guides/src/Nodes/Metadata/CopyRightNode.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @link https://phpdoc.org
+ */
+
+namespace phpDocumentor\Guides\Nodes\Metadata;
+
+/**
+ * The copyright element contains the document's copyright statement.
+ */
+class CopyRightNode extends MetadataNode
```
jaapio, line 19:
```suggestion
class CopyrightNode extends MetadataNode
```

packages/guides/src/Nodes/Metadata/AuthorsNode.php
```diff
@@ -0,0 +1,88 @@
...
+    /** @var AuthorNode[] */
+    private array $authorNodes;
+
+    /**
+     * @param Node[] $children
+     */
+    public function __construct(string $plaintext, array $children)
+    {
+        $this->authorNodes = [];
+        if (count($children) === 1) {
```
jaapio, line 38: If we introduce factories for the nodes, it would be nice to move this logic onto the factory. As this might only apply to RST?

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldListRule.php
```diff
@@ -0,0 +1,243 @@
...
+use RuntimeException;
+
+use function mb_strlen;
+use function mb_substr;
+use function preg_match;
+use function strtolower;
+use function trim;
+
+/**
+ * @link https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists
```
jaapio, line 47: This seems to be incorrect?

jaapio (changes requested):
Thanks for this improvement! It would be nice if we could move the creation of metadata nodes to configurable factories. This would allow people to introduce their own fields when needed.

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/RuleContainer.php
linawolf: Because field lists items that are parsed before the first title on a page is found create meta-nodes and are NOT output in the document. No FieldlistNode is created.

packages/guides/resources/template/html/guides/body/field-list.html.twig
```diff
@@ -0,0 +1,17 @@
+<table{% if node.classes %} class="{{ node.classesString }}"{% endif %}>
+    {% for definitionListTerm in node.children %}
```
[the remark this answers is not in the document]
linawolf: This is not a list. This is a field list, which is a different concept form a list

packages/guides/src/Nodes/FieldLists/FieldListItemNode.php
linawolf: A field list that is located before the page title is treated as metadata. It does not create a FieldListItemNode. A field-list later in the document is just output in a way defined by the project. Most HTML-based output is using tables here.

packages/guides/src/Nodes/Metadata/OrphanNode.php
linawolf: It is really a Metadata Node without content. But as an empty string is usually treated similar to false I just put another String?

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldListRule.php
linawolf: wupps

packages/guides/src/Nodes/Metadata/CopyRightNode.php
linawolf: done

packages/guides/src/Nodes/Metadata/AuthorsNode.php
linawolf: Should this be a factory or rather a production?

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldListRule.php
linawolf: Should this be factories or productions?

jaapio (approved):

PR #304, opened 2023-04-16
Remove strange `-` prefix for admonition classes

wouterj:
I guess this is a leftover from a class name like `admonition-tip`, but `-tip` as a class name doesn't make much sense to me. These classes are currently not used by phpDocumentor's stylesheets, so I think it's safe to fix this weird case now before more tooling relies on it :)


jaapio:
thanks :-)

PR #310, opened 2023-04-20
Correctly check if file exists in case insensitive file system

mabolek:
The code was checking whether a file exists or not, but using a negation. The code still seemed to work on case insensitive file system because we tested for "index" and "Index", and even though "index" didn't exist, failing would cause no issue because "Index" existed.

The issue became apparent when testing on case sensitive file systems (macOS).


jaapio:
Nice catch!
It would be very nice if we could add a unittest to cover this case.  Are you able to do that?

jaapio:
Merged, thanks, let's have a look at the tests in another PR

PR #330, opened 2023-04-21
Rebased: Add PreRenderDocument and PostRenderDocument events

linawolf:


packages/guides/src/Renderer/LatexRenderer.php
```diff
@@ -28,7 +31,7 @@ public function supports(string $outputFormat): bool
 
     public function render(RenderCommand $renderCommand): void
     {
-        $renderDocumentHandler = new RenderDocumentHandler($this->renderer);
+        $renderDocumentHandler = new RenderDocumentHandler($this->renderer, $this->eventDispatcher);
```
jaapio, line 34: This should also be removed as you did in the HtmlRenderer by using the commandbus.

jaapio (changes requested):
Thanks for addressing some bugs in this PR! 

Just a small improvement that should be done before we can merge this.

packages/guides/src/Renderer/LatexRenderer.php
linawolf: @jaapio I don't really get yet what the command bus is doing and how to use it.

jaapio: The commandbus is routing a command to a handler. 
Some background info:
https://barryvanveen.nl/articles/49-what-is-a-command-bus-and-why-should-you-use-it

jaapio:
Yeah, yet another step forward! Thanks

PR #333, opened 2023-04-21
[POC] Extensions & user configuration

wouterj:
Very early draft of my idea (when pursuing the idea of using the DI component) for extensions and user configuration.

There are a couple things I did:

* **Centralize the code building the container into a factory.** This makes it easier to change the build process (instead of having to update both the tests and `guides` "binary"), and allows to add some custom layer for DI extensions.
* **Give extensions full control** by allowing them to add compiler passes, create configuration trees and load services. This means the container initialization no longer has to register compiler passes that actually belong to a specific package. I've also moved the rst extension to the rst package, as that makes more sense to me when adding markdown later on.
* **Create a structure to configure extensions**, using the recently added `template_paths` for the main package as an example. For now, you configure extensions by calling `loadExtensionConfig()` on the container factory.
   In a future PR, we can add a file loader layer that can read e.g. a `guides.xml` config file in the user project.


jaapio:
I like this idea in general, moving the initialization of the container into a factory makes sense to me. I already had some issues because I forgot to update the cli app, or the tests. So having those in one location makes sense to me!

When I created this library at first and started to split things into separate packages I never thought about how to integrate it with an application. This was built on request, as I noticed that it was hard to get started with the package and a lot of configuration is needed to make it work. 

To reply here to your questions about phpDocumentor core, phpDocumentor doesn't have a working extension system, I did play with Symfony bundles a bit, as phpDocumentor is a normal Symfony application with a kernel and such. Maybe I will transform that at some point as the framework bundle is forcing me to use the HTTP kernel. Which I do not like, but I do like the concept of bundles. When I was building the initial version of the application here, I did have a look at how bundles work in Symfony, as I wanted to make the guides-symfony package a bundle, but without the HTTP stuff. So I do not have any hard feelings on bringing the guides-symfony and guides-cli closer to a Symfony console application if that will simplify the setup. 

The main reason for me to have a single location for the integration of Symfony was that I didn't want to introduce a dependency on symfony DI and configuration in the guides-core, guides-rst and guides-markdown. But if we do not add it as a composer dependency, and make sure it is never connected to other parts of the libraries, I think it is ok to move the DI code into those packages. 

**Extra**
From a phpDocumentor perspective, it would be nice if we could use the same extension logic to integrate the libraries, as I said phpDocumentor is now a symfony application, so I added the extensions to the kernel to get things loaded. If there was a way to use the factory for this, that would be very helpful. Right now I need to change things in phpDocumentor to get this library back to work. 

**Configuration** 
In phpDocumentor we are using cli options and config files to configure the runtime of the project, we keep this out of the actual container initialization, I think it would be nice to have the same separation here, extensions should of course be connected to the container, but input and output format are more runtime configuration things that can be switched. 
I'm not 100% sure if this is applicable here, The configuration options are limited, but as I see this repo moving more and more to a full application, I think it would be good to keep in mind that there might be a difference between runtime and "cachable" configuration that is not changing the runtime directly. like the output format  

I hope this answers your questions a bit, if not let me know, Then I will try to elaborate a bit more on this.

wouterj:
I think we can create a structure that works with both `guides-cli` and `phpdocumentor`, given they both rely on the Symfony DI container.

The extension introduced in this PR fully relies on interfaces known and supported by the DI component. So if `guides-rst` and `guides-core` have a DI extension, it should be relatively easy to use this one extension class to register it in both applications.

I'll continue to work on `guides-cli` with extension support (using a config file). Once that works, we can probably copy some logic to `phpdocumentor` (or create a shared package, but I'm not sure if that is great for maintainability) and also support registering guides as a phpdocumentor extension.

---

About configuration, I agree that we should differentiate between runtime and "static" config. Although they of course can overlap, e.g. we can configure some service defaults based on a config file, and allow to override these defaults using CLI options. This means we still have a cacheable container (using a `guides.xml` config file hash to invalidate the cache), but also override this using runtime config.

wouterj:
Continuing this in #388 and future PRs.

PR #335, opened 2023-04-22
Tag rules with priority

wouterj:
Relying on the order of registration is prone to errors and not really self-explanatory. Symfony comes with a little helper that allows to easily sort rules using a `priority` tag attribute. I've left some space between each priority to allow new rules to position themselves in between the existing ones.

I've configured the priority using constants for two reasons:

* This makes it easier to change priorities later on in a backwards compatible manner. E.g. if I want to add a custom rule in my application that is run before the link rule, I can use `LinkRule::PRIORITY + 1` and this library can increase the priority of the link rule by any value without breaking my application.
* This can also be used by applications not using the Symfony container configuration in the `guides-symfony` package.


jaapio:
Nice thanks!

PR #336, opened 2023-04-22
Improvement/directive usage

jaapio:
- changed the usage of directives to be able to use the model produced by the DirectiveRule in the directive handlers. This allows us to simplify the code and removes all redundant docblocks. 
- Centralize the variable set on documents when handling directives.


linawolf (approved):

linawolf:
thanks for this needed change

PR #337, opened 2023-04-23
Declare properties readonly if possible

derrabus:
* The first commit contains all occurrences that PhpStorm could fix automatically.
* The second commit contains changes I've applied manually.


linawolf (approved):

linawolf:
Thank you, I believe it is good practice to use readonly here.

Issue #307, opened 2023-04-17
Anchors before a new section should replace the implicit section ID


wouterj:
Consider this rst document:

```rst
Section A
---------

.. _custom-id:

Section B
---------
```

It should contain 2 sections: Section A with (implicit) id `section-a` and Section B with (explicit) id `custom-id`.
Currently, this library adds an `<a name="custom-id"></a>` in front of Section B, but still keeps its implicit ID (`section-b`).

We should "collapse" an `AnchorNode` if it's directly followed by a `SectionNode` using a compiler pass (#221 implemented a similar logic for the `.. class::` directive).

linawolf:
We solved this for the doctrine/rst-parser and should move the solution (if applicable) and the tests here

linawolf:
@wouterj do the tests in this PR cover what you need once they are solved?

https://github.com/phpDocumentor/guides/pull/308

wouterj:
Yes, thanks for adding them!

jaapio:
I'm going to try to summarize what I learned about references in RST... this is going to be a long story... as it is slightly complicated. But also diverse between standards...

We know 2 types of elements that can be used to create links. Doc utils names them as implicit and explicit hyperlink targets.
Implicit targets are elements that you can link to because of there type. Docutils has the following elements that may define implicite targets. titles footnotes citations

Explicit targets are the ones that we define as .. _target: optional followed by a link or other target.

To avoid conflicts  implicit links are overwritten by explicit ones... so in this construct:
```
.. _my-reference-label:

Section to cross-reference
--------------------------

This is the text of the section.

It refers to the section itself, see :ref:`my-reference-label`.
```
Without the explicit section we would have a link id with:
section-to-cross-reference

It is unclear to me, but not unexpected a link Section to cross-reference could work. However docutils does not specify the normalization of implicit link targets..

With the explicit .. _my-reference-label:  the implicit target is removed, according to the sphinx documentation. https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref

Multiple explicit link targets are allowed above those elements.
in html this will result into replacing the id of a title with the explicit link id. When multiple links are added, multiple anchors are added? (need to check this with sphinx)

---

Right now our implementation supports (with bugs) the text roles ref and doc the latter seems to be implemented invalid.... Docs should resolve to documents, and not to sections, it is some how supported by the implementation to link to sections. According to sphinx doc should only be used to link documents, not sections.

[...]

PR #338, opened 2023-04-23
Attach Anchor nodes to sections

linawolf:
I solved this in a way that anchor nodes can also be attached to other kind of nodes In rst anchors can be attached to figure notes as well, where the caption of the figure is treated as title. This is not implemented yet. Compare https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref

resolves https://github.com/phpDocumentor/guides/issues/307

### DIFF
--- a/packages/guides/src/Compiler/NodeTransformers/AnchorNodeTransformer.php
+++ b/packages/guides/src/Compiler/NodeTransformers/AnchorNodeTransformer.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Compiler\NodeTransformers;
+
+use phpDocumentor\Guides\Compiler\NodeTransformer;
+use phpDocumentor\Guides\Nodes\AnchorNode;
+use phpDocumentor\Guides\Nodes\DocumentNode;
+use phpDocumentor\Guides\Nodes\Node;
+use phpDocumentor\Guides\Nodes\TitledNode;
+
+/**
+ * @implements NodeTransformer<Node>
+ *
+ * A custom anchor, also called label, can be applied to sections if directly in front of a headline. It can be applied
+ * directly to a figure as well. In all other cases it is a standalone target rendered separately, which does not have
+ * a title.
+ *
+ * https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref
+ */
+class AnchorNodeTransformer implements NodeTransformer
+{
+    /** @var AnchorNode[] */
+    private array $anchorNodes = [];
+
+    public function enterNode(Node $node): Node
+    {
+        if ($node instanceof DocumentNode) {
+            // unset classes when entering the next document
+            $this->anchorNodes = [];
+        }
+
+        if ($node instanceof AnchorNode) {
+            // collect all anchors, multiple anchors can be applied to one section
+            $this->anchorNodes[] = $node;
+        } elseif ($node instanceof TitledNode) {
+            // Titled Nodes handle anchors themselves and provide a title to links
+            foreach ($this->anchorNodes as $anchorNode) {
+                $node->addAnchor((string) $anchorNode->getValue());
+                $anchorNode->setParentNode($node);
+                // reset the anchor nodes as they have already been added to a titled node
+                $this->anchorNodes = [];
+            }
+        }
+
+        return $node;
+    }
+
+    public function leaveNode(Node $node): Node|null
+    {
+        return $node;
+    }
+
+    public function supports(Node $node): bool
+    {
+        return true;
+    }
+}
--- a/packages/guides/src/Nodes/SectionNode.php
+++ b/packages/guides/src/Nodes/SectionNode.php
@@ -5,13 +5,26 @@ declare(strict_types=1);
 namespace phpDocumentor\Guides\Nodes;
 
 use function array_merge;
+use function array_unique;
+use function trim;
 
 /** @extends CompoundNode<Node> */
-final class SectionNode extends CompoundNode
+final class SectionNode extends CompoundNode implements TitledNode
 {
+    private string $identifier;
+    /** @var string[] */
+    private array $anchors = [];
+
     public function __construct(private readonly TitleNode $title)
     {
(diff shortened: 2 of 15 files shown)


packages/guides/src/Compiler/NodeTransformers/AnchorNodeTransformer.php
```diff
@@ -0,0 +1,59 @@
...
+                $anchorNode->setParentNode($node);
+                // reset the anchor nodes as they have already been added to a titled node
+                $this->anchorNodes = [];
+            }
+        }
+
+        return $node;
+    }
+
+    public function leaveNode(Node $node): Node|null
```
derrabus, line 50:
```suggestion
    public function leaveNode(Node $node): Node
```

The method can never return null.

linawolf: No, but it is declared like that in the parent class?

packages/guides/src/Compiler/NodeTransformers/AnchorNodeTransformer.php
```diff
@@ -0,0 +1,59 @@
...
+            $this->anchorNodes = [];
+        }
+
+        if ($node instanceof AnchorNode) {
+            // collect all anchors, multiple anchors can be applied to one section
+            $this->anchorNodes[] = $node;
+        } elseif ($node instanceof TitledNode) {
+            // Titled Nodes handle anchors themselves and provide a title to links
+            foreach ($this->anchorNodes as $anchorNode) {
+                $node->addAnchor((string) $anchorNode->getValue());
```
jaapio, line 40: I would like to suggest to add the anchor nodes as child nodes on the title. And remove the anchor from the tree. 

Like you relocate it in the tree... This will simplify the rendering, as we can just render a list of anchors before the parent.

We could introduce a method getAnchors() that will filter the anchors from the children.

The other benefit is that we could remove the getParent method, as this might complicate caching because of circulaire references.

packages/guides/src/Compiler/NodeTransformers/AnchorNodeTransformer.php
jaapio: If you apply my suggestion to relocate the anchors this method should return null, this will remove the anchor from it's current location.

packages/guides/src/Nodes/SectionNode.php
```diff
@@ -5,15 +5,28 @@
...
+    private array $anchors = [];
+
     public function __construct(private readonly TitleNode $title)
     {
+        $this->identifier = $title->getId();
+
         parent::__construct();
     }
 
+    public function getTitlePlaintext(): string
```
jaapio, line 25: I think this will not work, as the title will contain markup?

packages/guides/src/Nodes/SectionNode.php
```diff
@@ -5,15 +5,28 @@
...
-final class SectionNode extends CompoundNode
+final class SectionNode extends CompoundNode implements TitledNode
 {
+    private string $identifier;
+    /** @var string[] */
+    private array $anchors = [];
+
     public function __construct(private readonly TitleNode $title)
     {
+        $this->identifier = $title->getId();
```
jaapio, line 20: Why would you store this in a property? GetIdentifier could simply proxy to title?

packages/guides/src/Nodes/SectionNode.php
```diff
@@ -33,4 +46,22 @@ public function getTitles(): array
 
         return $titles;
     }
+
+    public function addAnchor(string $anchor): void
+    {
+        $this->identifier = trim($anchor);
```
jaapio, line 52: The last anchor is determining the identifier of the section?

jaapio (changes requested):
I think a node traverser might not be the way to or we do miss a mechanism in the compiler that would allow you to move forward to the next anchor related element. 
 
So if you find a anchor, start scanning for a title or a section and link it? 

Let's discuss this approach as I'm not fully aware of the features you might need in the compiler.

packages/guides/src/Nodes/SectionNode.php
linawolf: If no explicit anchor is given, the identifier is derived from a converted form of the title text. if one explicit anchor is given it is the identifier of the anchor directly following the section, therefore the anchor last found before the section

packages/guides/src/Nodes/SectionNode.php
linawolf: I think we should be able to derive a non-marked up plain text version of the title that can be used as link-text etc later? we do the same when we create the default identifier of the title?

packages/guides/src/Compiler/NodeTransformers/AnchorNodeTransformer.php
linawolf: Thing is, we cannot always relocate them. If the anchor is directly followed by a new section or a figure or any other titled node that supplies a title for a link target it is directly attached to that one. If any other Node comes next we need to keep it in place. It can be linked to but does not supply a link text.

I think that the section is a more logical location then the title because we are linking to a section, not just the title usually.

We could remove the getParent if that is problematic I guess. I figured we might be able to use it to relocate nodes in a secong preprocessing step if necessary. With the current tools we have I see no other way of manipulating the AST the way that would be needed. If we turned the nodes into a bidirectional graph that would allow us to traverse and manipulate it...

packages/guides/src/Compiler/NodeTransformers/AnchorNodeTransformer.php
linawolf: actually they cant be removed without looking forward as they might not be attached to a section.

jaapio:
@linawolf: https://github.com/phpDocumentor/guides/blob/main/packages/guides/src/Compiler/Passes/ImplicitHyperlinkTargetPass.php is this not the same feature?

linawolf:
I don't think so. 

a) It seems to only work on Document-Level so it does nothing to subsections
b) It seems to simply remove Anchor nodes without making a replacement

linawolf:
I think we will not get it finished like this

PR #339, opened 2023-04-25
Added skipped test for latex rendering

linawolf:
currently latex rendering uses HTML templates and outputs no latex.latex


jaapio:
Thanks for adding this!

PR #342, opened 2023-04-29
Introduce Literalinclude Directive

linawolf:


jaapio (commented):
Is this directive not an alias of code-block like this? If I remember it correctly, literal block should be a `<pre>` block and not a code block?

linawolf:
We are not talking about a literal block here. A literal **include** includes a file directly into a code-block

PR #343, opened 2023-04-29
Ignore Todo Directives

linawolf:
They should be treated as comment


jaapio (approved):
Question, should we collect the todo nodes in the ast and not render them? This would allow us to create a report of all to-do items in the docs. 

We could create a nooprenderer that accepts a list of nodes that are not rendered. 

This pr is still an improvement as it introduces some support for the todo directive and therefore I approve it. We can always iterate more on this.

PR #346, opened 2023-04-29
Some toctree improvements + add skipped test for non-titlesonly toctrees

wouterj:
This library has the same issue as doctrine/rst-parser, where each toctree is assumed to be `:titlesonly:` and no sub-section titles are rendered. Although we are not using it in the Symfony documentation, we do have a test for this in our test suite somehow.

Lina has started a PR in the past to fix this for the Doctrine parser (https://github.com/doctrine/rst-parser/pull/199), we might reuse the logic from that PR to fix this test in the future.

Meanwhile, I've already made some minor improvements:

* Use relative paths as test name, this makes it easy to filter data sets (e.g. `./vendor/bin/phpunit tests/Integration/IntegrationTest.php --filter testHtmlIntegration@toctree`, currently "toctree" must be the full local path)
* Remove the `spaceless` tag from the template. This tag is deprecated in Twig as has quite some edge cases where meaningfull spaces are removed, instead we should rely solely on the [whitespace control characters](https://twig.symfony.com/doc/3.x/templates.html#whitespace-control). As a benefit, it makes the diff from failing toctree tests much easier to read (as each item is on its own line now)

### DIFF
--- a/packages/guides/resources/template/html/guides/body/toc/toc-item.html.twig
+++ b/packages/guides/resources/template/html/guides/body/toc/toc-item.html.twig
@@ -1,9 +1,8 @@
-<li class="toc-item">
-    <a href="{{ url }}">{{ renderNode(node.value.value) }}</a>
+<li class="toc-item"><a href="{{ url }}">{{ renderNode(node.value.value) }}</a>
 
-    {% if node.children|length %}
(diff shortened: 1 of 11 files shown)


jaapio:
Thanks for these improvements

PR #347, opened 2023-04-30
Introduce Contents Directive

linawolf:


packages/guides-restructured-text/src/RestructuredText/Directives/ContentsDirective.php
```diff
@@ -0,0 +1,43 @@
...
+        DocumentParserContext $documentParserContext,
+        Directive $directive,
+    ): Node|null {
+        $options = $directive->getOptions();
+        $absoluteUrl = $this->urlGenerator->absoluteUrl(
+            $documentParserContext->getContext()->getDirName(),
+            $documentParserContext->getContext()->getCurrentFileName(),
+        );
+
+        return (new TocNode([$absoluteUrl]))
```
jaapio, line 40: Should this be a TocNode or a contentnode that you introduced in this pr? 

If you use the contentnode, you might need to adjust the compiler pass so it accepts the content node. 
It might be worth it to us the contents node so people can manipulate the TOC and contents directive result differently.

jaapio (commented):
Thanks for adding this, just a question about the used node, appart from that is looks good 👍

packages/guides-restructured-text/src/RestructuredText/Directives/ContentsDirective.php
linawolf: Yeah, I was thinking about this as well. But is there such a concept of toc / contents outside of rst?

jaapio: If we introduce a contents node, one could use it to create a compilerpass to create a content node based on the heading of a document. Even if the markup language doesn't support it.

There are some extensions on markdown to support a table of contents, like sphinx does for RST. As the original format was not designed to have such feature. 

The commonmark library we are planning to use for markdown has such extension.

linawolf: I gave it another try to unify the menu handling

jaapio:
Thanks for your retry! This is great

PR #348, opened 2023-05-02
[TASK] Add example tests

tomwarwick:


linawolf:
@tomwarwick thanks a lot for your contribution!

These tests do not work yet - which is to be expected. So we need to disable them by adding a file called "skip" to the input directory. Have a look at how we are doing it for this test, for example: https://github.com/phpDocumentor/guides/tree/main/tests/Integration/tests/anchor-title-overriden-by-ref-title

Once the underlying issue or missing feature is fixed we will then remove the file called "skip"

jaapio:
@tomwarwick I'm not sure what happened but it looks like something went wrong in your git history, I would recommend you to create a new branch from main and cherry-pick your commits:

`git cherry-pick cc7b86c`
`git cherry-pick 8778cdf `

Please let me know when  you need some help.

linawolf:
@tomwarwick I cherry picked your changes into here: https://github.com/phpDocumentor/guides/pull/387 so we can get them merged.

PR #351, opened 2023-05-06
Split themes and make phpdocumentor theme optional

linawolf:
- Make read the docs standard as suggested by @wouterj https://sphinx-rtd-theme.readthedocs.io/en/stable/
- Make the phpdocumentor theme a separate package

Test theme packages on the example of Admonitions


wouterj:
I have been working on the rtd theme in my repo: https://github.com/wouterj/symfony-docs-guides/tree/main/lib/docs-builder/templates/rtd

As far as I could see with the Symfony docs, those are all necessary changes to have a full rtd theme (it also implements code highlighting, something that is not build in in this package, so the `code.html.twig` file needs some changes).

wouterj:
Btw, given the phpDocumentor and rtd HTML is so similar, I can also think of a situation where we slightly update the CSS from phpDocumentor and ship that instead of the rtd CSS. As long as the HTML is generic and doesn't contain any `phpdocumentor-...` stuff, I think having your own default theme is a good "marketing" strategy.

linawolf:
As far as I understand @mvriel the php documentor theme needs to be plain with no complete page, no styles etc as it is then integrated as webcomponents. On the other hand in a standalone version we do want a complete site as it seems you are doing in https://github.com/wouterj/symfony-docs-guides/blob/main/lib/docs-builder/templates/rtd/html/guides/structure/layout.html.twig?

wouterj:
Ah, I once got a screenshot from Jaap with a phpDocumentor typed rendering of the Symfony docs. That made me think this package shipped a full theme as well. But that's all I know about rendering in the phpDocumentor process, so it could very well be that this package ships plain components and phpDocumentor turns it into a nice page.
Let's see what Mike or Jaap think then :)

jaapio:
my intension with this project and all related projects was to provide a solid base for people to create there own themes. The phpDocumentor specific theme should be part of phpDocumentor core. 

Since Lina is co-maintaining this project and suggested introducing some theme system in the project, it is ok for me to have a demo theme that uses the clean blocks from the core with some extensions to make it fit in Bootstrap. But from my point of view, I do not have much interest in pushing hard on themes. phpDocumentor itself has its own template system and rendering works a bit differently than what this library provides. 

Mike suggested keeping the layout template clean and not impacting the components (node templates) too much on introducing the bootstrap theme as phpDocumentor theme relies on them. We try to keep CSS and HTML together in a webcomponent like style. But it is definitely no web components. 

----

What I showed to @wouterj months ago was the output generated by phpDocumentor, using the phpDocumentor theme. Which is shipped with phpDocumentor itself. Since this is closely build in phpDocumentor's rendering I would not suggest to extract it, This library should just focus on building blocks. Not on full blown themes, the new bootstrap theme could do that. To make it easier for people to get started with it. 

--- 

From a phpDocumentor point of view, we prefer people to use phpDocumentor as application, not the cli component shipped in this repo :-) for me the cli tool was more a test case to help others to get started with this project, and start extending it. I do not have time nor the interest to build 2 applications. I don't want to push anybody to use phpDocumentor, and I do understand that people have there own usecases for this library. So if we can keep the cli tool as a bare minimum to run this library as a standalone tool, it is ok for me to keep it that way. 

---

I think Typo3 and Symfony will have their own extensions on the cli tool in the end, if they will use the tool at all. They don't need much configuration and customization at runtime, it's just a tool that renders docs, every time in the same way. So if we provide the building blocks for the projects to build something, maybe with a small application that can be extended easily it will work for everyone, but not for phpDocumentor itself.

jaapio (approved):
The changes in this pr do not conflict with my intentions with this repository. This is good to go. 

Thanks for your changes @linawolf!

PR #356, opened 2023-05-06
Fail incomplete tests that pass

wouterj:
As we're using incomplete tests as a living todo list, I think it's good to also check whether such tests still fail.

This PR changes the functional and integration test to always run the tests and mark as incomplete only if there are errors. If there are no errors, but the test is skipped, it marks the test as failed.

As you see, this already discovered one skipped test that we have fixed in the past months :)


linawolf:
Wow, great Idea!

linawolf (approved):

PR #357, opened 2023-05-06
 Run Bootstrap Theme as Application

linawolf:
Not sure if this is the best way to go.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
         "phpdocumentor/guides-graphs": "dev-main@dev",
         "phpdocumentor/guides-restructured-text": "dev-main@dev",
         "phpdocumentor/guides-symfony": "dev-main@dev",
+        "phpdocumentor/guides-theme-bootstrap": "@dev",
         "psr/event-dispatcher": "^1.0",
         "psr/log": "^1.1",
         "symfony/contracts": "^2.5",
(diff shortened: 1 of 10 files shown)


wouterj:
I think we should only keep one application (`guides-cli`) and add extension support to it. Themes would then ship an `Extension` PHP class that configures a theme in the DI container.

In a CLI option (or config file), a user can then specify which theme it wants. Something like:

```php

class ThemeManager
{
    private array $themeMap = [];

    public function __construct(
        private \Twig\Loader\FilesystemLoader $loader,
        private array $defaultPaths
    ) {
    }

    public function registerTheme(string $name, array $paths)
    {
        $this->themeMap[$name] = $paths;
    }

    public function useTheme(string $name)
    {
        $this->loader->setPaths(array_merge($this->defaultPaths, $this->themeMap[$name] ?? []));
    }
}
```

linawolf:
That sounds like a better idea, however I don't really know how to build it :(

wouterj:
Yeah, the concrete implementation is also not 100% clear in my head. In https://github.com/phpDocumentor/guides/pull/333 , I'm trying to get extensions working that we can then apply for the bootstrap theme. I'll keep you posted on that experiment :)

linawolf:
I will work on the ThemeManager idea

PR #359, opened 2023-05-07
Overhaul document (meta) title handling

linawolf:
In HTML each document must have exactly one <title> tag. In general it makes
sense to have a unique title in non-HTML documents as well. This title is the first title found in a documents body, overridden by .. title:: meta directive. It is considered good practice to also attach the project title for display in the <title> tag in HTML.

The project title is a concept also needed in other areas like the intersphinx linking. TODO: Make the project title configuratble by project settings - to be introduced.


jaapio (changes requested):
I think we are reaching the a point where project settings are needed before we continue with this. The ProjectMeta class seems to be a starting point for this, except for the fact that it is part of the container. 

The DI container should be cached in the end, right now the metas object is violating this approach, and it is kind of hard to get rid of it, But I want to do that at some point. All runtime configuration should be in the parser and rendering context, with an additional configuration object that is shared over the parsing and rendering stage. I suggest that you move the projectMeta into the run command and add it to the RenderCommand so it is accessible when the renderContext is created. And keep it out of the DI container. This will allow you to set the title of your project as a CLI option, and later when we introduce something like project settings we could load it in the run command as well, and populate the projectMeta with it. 

**Suggestion**: Should we rename `ProjectMeta` to `ProjectSettings`? 

I do like the changes you did to have a single documentTitle! 

Thanks!

linawolf:
This data cannot be only changed by settings but also by meta field-lists, see https://github.com/phpDocumentor/guides/pull/360
Therefore the naming

linawolf:
@jaapio the value in DI is more like a default value that you should change via field list or settings

jaapio (approved):
This looks great! thanks!

PR #367, opened 2023-05-13
Use section title as default reference title

linawolf:
depend on #366


jaapio:
I think this is good to be merged, could you please rebase this pr so I could do a final check?

PR #368, opened 2023-05-13
Introduce project settings

linawolf:


jaapio (changes requested):
Thanks for this addition. Introducing projectsettings makes a lot of sense, but I think we should keep them separate 
from the `ProjectMeta`. You could just directly assign the required fields from the projectsettings in the projectmeta
in the run command, can't you?

linawolf:
> Introducing projectsettings makes a lot of sense, but I think we should keep them separate from the `ProjectMeta`.

How exactly would you do that?

linawolf:
@jaapio I am not sure how to proceed here, please clarify

jaapio:
ProjectSettings is for me a collection of settings defined by the user when running the application. That means that `input-format` and `output-format` are part of the settings in my opinion so is the theme. I compare this to the settings phpdoc and PHPUnit define in XML. Those settings allow users to define their configuration in code and commit it to their project. 

ProjectMeta is another layer of configuration, defined in the RST format. And we need to merge those at some point. ProjectMeta is a result of the parser, together with the parsed documents? So if we put the ProjectMeta in the context of the parser, we could merge those in an eventlistener (yes events need to be adjusted for this).

So steps would be:
- start application
- read ProjectSettings from file
- configure the runtime (input, output, theme, title) 
- pre parser get the ProjectMeta from the ParserContext and overwrite the title & version with ProjectSettings
- run the parser
- run compiler
- push the ProjectMeta in the renderContext
- start rendering

During the rendering, we have the renderContext to carry the ProjectMeta. That allows us to do the title rendering etc.

---

I'm not sure if I'm now overcomplicating stuff, and if this makes it more clear for you. Please let me know, otherwise, we can schedule a call to talk about this.

linawolf:
I believe you are thinking about a different kind of settings here, like about the ones that wouter introduced. These settings should replace the settings.cfg of Sphinx. This is information that is needed for rendering, parsing or in the passes. In Sprinx it is done like this: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/blob/main/Documentation/Settings.cfg It contains information about project (output in rendering) edit on github button (rendering) intersphinx mapping (already needed in the passes) and can in some cases be used to define additional textroles (we might not want to support that, but it would be needed during parsing). The settings are part of the data of the project / document tree. different versions of the project might have different settings. However I would not include technical information like which Extensions are loaded or what are the paths in here. As this information is not part of the project itself

jaapio:
I'm getting confused by the number of different settings in this situation. That means that we will have 4 levels of settings?
- guides.xml
- cli arguments & options
- ProjectSettings
- ProjectMeta

The first 2 are application settings? where the last 2 are project settings? Maybe we should name them as such, remove the ProjectMeta and directly configure the ProjectSettings when reading fieldset lists? 
I'm trying to avoid too many layers that people won't understand.

linawolf:
I am starting starting to wonder if ProjectMeta should be a Node. We would have a ProjectNode, it knows its DocumentNodes and stores Project wide information

jaapio:
Interesting... that would also remove the need of metas? As we can store them in the projectNode?

linawolf:
Should we give it a try? However without a Node being mutable that will be a real hassle, as we would have to change the ProjectNode whenever we find a new link or some such

jaapio:
Maybe we should make an exception, however... the immutable behavior is not a hard requirement for me it think. So if it becomes a real issue, we might need to consider removing it. 

Let's first give it a try with immutable, and if it becomes a problem let's see how we can change this.

linawolf:
The ProjectNode should be known from any DocumentNode. That means for each and every change we would have to inform all documents?

jaapio (changes requested):
Nice, this looks better. I still have a few questions.

I do not understand why the Project node is part of the DocumentNode, I would expect it to be the other way around. So a project is a collection of DocumentNodes. With the suggestion below you could still access the project from the DocumentParserContext.
If we put the ProjectNode in the RenderContext we have access to it during the rendering as well.

Your idea of introducing a ProjectNode is great, when you look at the phpDocumentor Reflection component you will notice that we are doing something equal with PHP files. We are adding them to a Project object. Which is the result of the Parser.

$projectFiles = [new \phpDocumentor\Reflection\File\LocalFile('tests/example.file.php')];
$project = $projectFactory->create('My Project', $projectFiles);

As suggested on Slack, it might be good to align the Rule interface with FieldListItemRule.

I think this might work:

interface FieldListItemRule extends Rule
{
    public function applies(DocumentParserContext $documentParser, FieldListItemNode $fieldListItemNode = null): bool;

    /** @param FieldListItemNode $fieldListItemNode */
    public function apply(DocumentParserContext $documentParserContext, CompoundNode $on = null): MetadataNode|null;
}

Thank you for this very nice improvement!

wouterj:
I didn't follow the whole discussion, but I'm not very convinced about the need for a `settings.php` next to `guides.xml`. I think we can combine them (which I'm also used to do with Sphinx's [`conf.py`](https://github.com/symfony/symfony-docs/blob/3.4/_build/conf.py) file).

1. We can allow extension configuration (the thing we now exclusively use with `prependExtensionConfig()`) to be included in the XML config:
   ```xml
   <guides>
     <extension class="phpDocumentor\Guides\Guides"
       title="Project title"
     />
   </guides>
   ```
   (we can do it so that the `GuidesExtension` can be configured through the root `<guides>` node)
2. In the `GuidesExtension`, we would preconfigure `ProjectSettings` in the DI container based on the config we receive.
3. Then, for some options we can allow CLI options to e.g. call `$projectSettings->setName($input->getOption('project-name'))` to override the default config values.

If we, in the future, add container caching we can check freshes based on `guides.xml`: As long as it is not updated, the container is still fresh. If it's updated, we have to rebuild the container.

jaapio:
From a phpDocumentor point of view, refreshing might become an issue, as we are already using the container cache. What you see in general is that people install phpDocumentor globally and then start parsing their projects. Having a shared cache with multiple projects might become an issue when using the configuration options of Symfony?

linawolf:
In TYPO3 we also work with a settings.cfg which is part of the data, not part of the Application. We use the same Application for a great number of input. Each input has to have its own settings

linawolf:
@jaapio I gave it another try, attached the project Node to the ParserContext/RenderContext and introduced a CompilerContext to also have it available there

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldList/AbstractFieldListItemRule.php
```diff
@@ -17,7 +19,7 @@ public function applies(FieldListItemNode $fieldListItemNode): bool
         return strtolower($fieldListItemNode->getTerm()) === 'abstract';
     }
 
-    public function apply(FieldListItemNode $fieldListItemNode): MetadataNode
+    public function apply(FieldListItemNode $fieldListItemNode, DocumentNode $documentNode, DocumentParserContext $documentParserContext): MetadataNode
```
jaapio, line 22: You can get the Document From the DocumentParserContext. There is no need to add is separately.

packages/guides/src/Compiler/CompilerContext.php
```diff
@@ -0,0 +1,29 @@
...
+ * @link https://phpdoc.org
+ */
+
+namespace phpDocumentor\Guides\Compiler;
+
+use phpDocumentor\Guides\Nodes\ProjectNode;
+
+class CompilerContext
+{
+    public function __construct(
```
jaapio, line 20: Now we have a Compiler context, maybe we should move the document in here as well.

linawolf:
@jaapio thanks for the suggestions, done!

packages/guides/resources/template/html/structure/layout.html.twig
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-    <title>{{ title }}{% if title!=null and guides_project.getTitle()!=null %} - {% endif %}{% if guides_project.getTitle() %}{{ guides_project.getTitle() }}{% endif %}</title>
+    <title>{{ title }}{% if title!=null and project.getTitle()!=null %} - {% endif %}{% if project.getTitle() %}{{ project.getTitle() }}{% endif %}</title>
```
jaapio, line 4: Can we now get this from the context? The `project` variable is conflicting with phpdocumentor core. That's why I renamed it. The context is available as `env` 

Extracting it from the context will help use to reduce the chance of conflicts.

packages/guides/src/Compiler/NodeTransformers/FootNodeNamedTransformer.php
```diff
@@ -19,11 +20,15 @@ public function __construct(
     ) {
     }
 
-    public function enterNode(Node $node, DocumentNode $documentNode): Node
+    public function enterNode(Node $node, CompilerContext $compilerContext): Node
     {
+        if ($compilerContext->getDocumentNode() === null) {
```
jaapio, line 25: I think it is good to move this test to the `getDocumentNode` method, to reduce the duplication of code.

jaapio (commented):
Do we still need the SettingsManager? If an extension wants to change the project settings, this can be done via a compiler pass or pre-rendering.

linawolf:
> Do we still need the SettingsManager? If an extension wants to change the project settings, this can be done via a compiler pass or pre-rendering.

The settings do not come from an extension but from the project data. So yes we still need them and more settings then just the project name will come once I get this finally merged. I wanted to keep this PR short and simple and it is starting to turn into a monster :(

packages/guides/resources/template/html/structure/layout.html.twig
linawolf: The project node is part of the data, not of the enviroment in my eyes?

jaapio: If you take it from the context, you can access it everywhere you might need it. Otherwise you have to add it in the renderers that need it, but get it from that same context?

jaapio: But please keep the name guides_project to prevent issues in phpDocumentor 😁

linawolf:
@jaapio done

jaapio (approved):

jaapio:
Merged! Thanks for all the improvements

PR #369, opened 2023-05-13
Output logs

linawolf:
this enables us, to test for warnings and errors in tests, it also improves troubleshooting for manual owners.

### DIFF
--- a/packages/guides-cli/src/Command/Run.php
+++ b/packages/guides-cli/src/Command/Run.php
@@ -8,6 +8,9 @@ use Flyfinder\Finder;
 use League\Flysystem\Adapter\Local;
 use League\Flysystem\Filesystem;
 use League\Tactician\CommandBus;
+use Monolog\Handler\ErrorLogHandler;
+use Monolog\Handler\StreamHandler;
+use Monolog\Logger;
 use phpDocumentor\Guides\Handlers\CompileDocumentsCommand;
 use phpDocumentor\Guides\Handlers\ParseDirectoryCommand;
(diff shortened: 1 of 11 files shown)


packages/guides-cli/src/Command/Run.php
```diff
@@ -76,6 +77,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $outputDir = $this->getAbsolutePath((string) ($input->getArgument('output') ?? ''));
         $sourceFileSystem = new Filesystem(new Local($input->getArgument('input')));
         $sourceFileSystem->addPlugin(new Finder());
+
+        $this->commandBus->handle(new LogToFileCommand($outputDir . '/logs'));
```
jaapio, line 81: I don't think this should be done via the command bus. The commandbus integration is only for parts of the library that need to be executed in a certain way. Monolog is not part of the core project of guides, we allow people to inject their own logger, So we can do this directly using monolog in the cli project. As that is an application, ready to use for end-users. 

Doing it via the command bus makes is more part of the core guides library, which I think is not a good idea. 

My suggestion is to make an input option to set the log path. which defaults to PHP://stderr.

linawolf: done, please have another look

jaapio (approved):

jaapio:
can you please rebase this PR :-) 

This looks very nice! Thank you!

linawolf:
@jaapio rebased, please merge

PR #370, opened 2023-05-13
Preserve whitespace and colon when textrole is preceded by common colon

linawolf:


packages/guides-restructured-text/src/RestructuredText/Span/SpanParser.php
```diff
@@ -327,12 +328,21 @@ private function parseInterpretedText(): string
                     }
 
                     $inText = true;
+                    break;
+                case SpanLexer::WHITESPACE:
+                    // textroles may not contain whitespace, we are not in a textrole but have found a common colon
+                    if ($inText) {
+                        $part .= $token->value;
+                    } else {
+                        $rollback = true;
```
jaapio, line 337: You can use `break 2` to break out of 2 levels. that will simplify the code a bit.

jaapio (changes requested):
Can you please update the unittests for this change, the spanparser test might need some extra cases.

linawolf:
@jaapio rebased, applied suggested changes, added unit tests. Please have another look

PR #371, opened 2023-05-14
Add new compiler pass to move anchors

jaapio:
Anchors should be part of the first compound node after their definition. Anchors at the end of a compound node are positioned wrong by the parser So they are moved to the next compoundnode.

### DIFF
--- a/packages/guides/src/Compiler/DocumentNodeTraverser.php
+++ b/packages/guides/src/Compiler/DocumentNodeTraverser.php
@@ -50,7 +50,7 @@ final class DocumentNodeTraverser
         }
 
         foreach ($shadowNode->getChildren() as $shadowChild) {
-            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext->withShadowTree($shadowNode));
+            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext->withShadowTree($shadowChild));
         }
 
         if (!$supports) {
--- a/packages/guides/src/Compiler/DocumentNodeTraverser.php
+++ b/packages/guides/src/Compiler/DocumentNodeTraverser.php
@@ -50,7 +50,7 @@ final class DocumentNodeTraverser
         }
 
         foreach ($shadowNode->getChildren() as $shadowChild) {
-            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext);
+            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext->withShadowTree($shadowNode));
         }
 
         if (!$supports) {
--- a/packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
+++ b/packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
@@ -27,7 +27,7 @@ final class MoveAnchorTransformer implements NodeTransformer
         }
 
         $this->seen[spl_object_hash($node)] = spl_object_hash($node);
-        $parent = $compilerContext->getShadowTree();
+        $parent = $compilerContext->getShadowTree()->getParent();
         $position = $parent->findPosition($node);
         if ($position === null) {
             throw new LogicException('Node not found in shadow tree');
--- a/packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
+++ b/packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
@@ -12,14 +12,18 @@ use phpDocumentor\Guides\Compiler\ShadowTree\TreeNode;
 use phpDocumentor\Guides\Nodes\AnchorNode;
 use phpDocumentor\Guides\Nodes\Node;
 use phpDocumentor\Guides\Nodes\SectionNode;
-
-use function spl_object_hash;
+use WeakMap;
 
 /** @implements NodeTransformer<AnchorNode> */
 final class MoveAnchorTransformer implements NodeTransformer
 {
-    /** @var array<string, string> */
-    private array $seen = [];
+    /** @var WeakMap<AnchorNode, true> */
+    private WeakMap $seen;
+
+    public function __construct()
+    {
+        $this->seen = new WeakMap();
+    }
 
     public function enterNode(Node $node, CompilerContext $compilerContext): Node
     {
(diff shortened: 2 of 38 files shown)


packages/guides/src/Compiler/Passes/MoveAnchorPass.php
```diff
@@ -0,0 +1,69 @@
...
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Compiler\Passes;
+
+use phpDocumentor\Guides\Compiler\CompilerPass;
+use phpDocumentor\Guides\Nodes\AnchorNode;
+use phpDocumentor\Guides\Nodes\CompoundNode;
+use phpDocumentor\Guides\Nodes\DocumentNode;
+
+class MoveAnchorPass implements CompilerPass
```
wouterj, line 12: Shouldn't this be a node transformer instead? That already has all code to traverse the tree of nodes.

The way I differentiate between them is:
* Node transformers operate within a single document (and therefor can even be cached)
* Compiler passes operate in the project, managing cross document logic (like toctrees).

jaapio: I think the current node tranformer cannot handle this, as I need to move elements. But I could be wrong... 
I have to make sure that I move elements at the right level, and in the node transformer, I don't know that?

wouterj: Hmm, maybe I have a wrong idea about the AST that we are dealing with here. But if it looks like this:

```
COMPOUND
    ...
ANCHOR
ANCHOR
COMPOUND
    ...
```

I think a transformer that supports compound and anchor nodes should work. It'll then store the anchors in a property when it discovers them, and push them in the compound node when it reaches one (clearing the property).

jaapio: I will check if it is possible to do this in a transformer, I think the main issue I will have is that the transformer will first handle the child nodes an then go to the next node. And its state is shared between the nodes. 

ANCHOR
COMPOUND
    ANCHOR
    ANCHOR
    COMPOUND
    ANCHOR
COMPOUND

Should be come

COMPOUND
    ANCHOR
    COMPOUND
    ANCHOR
    ANCHOR
COMPOUND
    ANCHOR

packages/guides/src/Compiler/DocumentNodeTraverser.php
```diff
@@ -50,7 +50,7 @@ private function traverseForTransformer(
         }
 
         foreach ($shadowNode->getChildren() as $shadowChild) {
-            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext);
+            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext->withShadowTree($shadowNode));
```
jaapio, line 53: This was a 🐛, the node transformer should get the current node as shadow element.

packages/guides/src/Compiler/DocumentNodeTraverser.php
```diff
@@ -50,7 +50,7 @@ private function traverseForTransformer(
         }
 
         foreach ($shadowNode->getChildren() as $shadowChild) {
-            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext);
+            $this->traverseForTransformer($transformer, $shadowChild, $compilerContext->withShadowTree($shadowChild));
```
jaapio, line 53: This was a bug, we should pass the current node in the shadow tree. So we can move around at the correct level.

packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
```diff
@@ -0,0 +1,105 @@
...
+use phpDocumentor\Guides\Nodes\Node;
+use phpDocumentor\Guides\Nodes\SectionNode;
+
+use function spl_object_hash;
+
+/** @implements NodeTransformer<AnchorNode> */
+final class MoveAnchorTransformer implements NodeTransformer
+{
+    /** @var array<string, string> */
+    private array $seen = [];
```
greg0ire, line 22: Should this be a WeakMap? So that you don't have to deal with hashes?

packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public function leaveNode(Node $node, CompilerContext $compilerContext): Node|null
+    {
+        //When exists in seen, it means that the node has already been processed. Ignore it.
+        if (isset($this->seen[spl_object_hash($node)])) {
+            return $node;
+        }
+
+        $this->seen[spl_object_hash($node)] = spl_object_hash($node);
```
greg0ire, line 36: If the value is unused, you might as well use `true`, `null` or 42.

packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public function supports(Node $node): bool
+    {
+        return $node instanceof AnchorNode;
+    }
+
+    public function getPriority(): int
+    {
+        return 30000;
```
greg0ire, line 57:
```suggestion
        return 30_000;
```

packages/guides/src/Compiler/ShadowTree/TreeNode.php
```diff
@@ -123,6 +124,18 @@ public function addChild(Node $child): void
         $this->node->addChildNode($child);
     }
 
+    public function pushChild(Node $child): void
+    {
+        if ($this->node instanceof CompoundNode === false) {
```
greg0ire, line 129: No need for strict comparison, we know this is a boolean

```suggestion
        if (! $this->node instanceof CompoundNode) {
```

packages/guides/tests/unit/Compiler/NodeTransformers/MoveAnchorTransformerTest.php
```diff
@@ -0,0 +1,177 @@
...
+                //phpstan:ignore-next-line
+                yield new MoveAnchorTransformer();
+            }
+
+            /** @return array<string, int> */
+            public function getPriorities(): array
+            {
+                return [];
+            }
+        }, 30000);
```
greg0ire, line 36:
```suggestion
        }, 30_000);
```

jaapio: Thanks, I added this as a style rule to php-code-sniffer

packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
jaapio: Thanks, I added this as a style rule to php-code-sniffer

packages/guides/src/Compiler/NodeTransformers/MoveAnchorTransformer.php
jaapio: I never used WeakMap, but it seems to be a very good use case here!

packages/guides/src/Compiler/ShadowTree/TreeNode.php
greg0ire: The exclamation mark might be missed indeed.

PR #383, opened 2023-05-19
Error handling for simple tables

linawolf:
There are probably more things to be tested, however this fixes the skipped test for now


jaapio:
Thanks!

PR #388, opened 2023-05-19
Introduce ContainerFactory

wouterj:
This is a split from #333 to try to make smaller steps more quickly :)

Goals of this feature from that PR description:

> * **Centralize the code building the container into a factory.** This makes it easier to change the build process (instead of having to update both the tests and guides "binary"), and allows to add some custom layer for DI extensions.
> * **Give extensions full control** by allowing them to add compiler passes, create configuration trees and load services. This means the container initialization no longer has to register compiler passes that actually belong to a specific package.


linawolf (approved):

linawolf:
This looks good! Thanks a lot!

jaapio (approved):

PR #389, opened 2023-05-20
Decouple the separate packages more

wouterj:
This includes #388 until it's merged. You can view the diff from this PR at https://github.com/phpDocumentor/guides/pull/389/files/1d10cbd81d28c6507dc34963892126c5264ab7aa..e0303fa756d17123defac13fc814571f4b736067

This removes some coupling between package resources:

* **The DI extensions and config are moved to the respective packages** (generic guides config to guides, rst config to rst). We don't need to depend on the DI component from these packages, as you can still use them without.
* **ReStructuredText specific templates are moved to the rst package**. By using prepended config, a package can now register extra base templating paths that will always be loaded.
* **Merged the remaining classes from guides-symfony into guides-cli**. With the extensions moved to the separate packages, you no longer need guides-symfony to plug guides and guides-rst into your own DI container (for phpDocumentor and the docs-builder). What was left in guides-symfony was some plumbing needed for the CLI package, which is why I think it makes more sense to move it there.

I've also refactored the `Configuration` class into the Symfony Config component. The main reason for this is that we need logic to merge multiple config resources (in this case, the prepended config from other packages, and maybe also user config in the future). The Config component has all logic we need to handle this merging + validation.
This config is separate from the project settings that @linawolf is introducing in #368. The config defined by extensions is compile time, while the project settings are runtime. We can try to make extension config configure a "default" project settings that can be overridden at runtime, but for now I would like to see them as 2 different things.

In a future PR, we can also add a config option for custom themes and start using that in the bootstrap theme package.


linawolf:
@wouterj some tests are still failing, please run the pre-commit tests locally

`make pre-commit-test`

jaapio:
Thanks @wouterj, this makes sense to me. 

I think we can discuss the way of configuration a follow up pr.

Thanks @linawolf  for the first review!

Issue #178, opened 2023-02-06
Add text role implementations


jaapio:
https://docutils.sourceforge.io/docs/ref/rst/roles.html

linawolf:
I am wondering if we can somehow use the work I put into https://github.com/doctrine/rst-parser/pull/257 to solve this

jaapio:
I think we can use parts of it, I replaced the full SpanParser to be able to split the parser into a lexing and parsing part. 

The lexer will break the text into smaller elements, called Tokens. The parser will consume those tokens to turn them into what I now named `SpanToken` But a better name would be `InlineNodes` as that refers to the inline markup that is supported by RST. 

I would like to introduce more of these kinds of `SpanToken`, the text roles you created could be just that! We could extend the rendering of the span to transform it completely into what we have now for nodes. So we can use the same mechanism as we have for other node types.

Issue #305, opened 2023-04-16
Text roles and whitespace control


wouterj:
Contrary to body elements, preserving original whitespace is important for inline elements.

Some [tests on the Symfony test repository](https://github.com/wouterj/symfony-docs-guides/actions/runs/4713936115/jobs/8359959701) are failing because of <s>missing spaces before an inline `<code>` element, as well as</s> new lines after a `</code>` tag. (EDIT: missing space before is probably due to a bit too heavy whitespace removal by the tests)

An inline newline is converted to a space in HTML, which works out most of the time but not when e.g. you put an inline literal before a final stop or comma (like at the end of the test case below).

See e.g. this test output:

```
3) SymfonyDocsBuilder\Tests\HtmlIntegrationTest::testBlocks with data set "text-roles/reference-and-code.rst" ('/home/runner/work/symfony-doc...de.rst', '/home/runner/work/symfony-doc...e.html')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<p>Next, create the template used to render the field in the<code translate="no" class="notranslate">index</code>and<code translate="no" class="notranslate">detail</code>\n
-<a href="https://symfony.com/doc/current/bundles/EasyAdminBundle/crud.html" class="reference internal">CRUD pages</a>.</p>\n
+'<p>Next, create the template used to render the field in the<code translate="no" class="notranslate">index</code>\n
+and<code translate="no" class="notranslate">detail</code><a href="https://symfony.com/doc/current/bundles/EasyAdminBundle/crud.html" class="reference internal">CRUD pages</a>\n
+.</p>\n
 <p>More about CRUD pages.</p>'

/home/runner/work/symfony-docs-guides/symfony-docs-guides/lib/docs-builder/tests/integration/HtmlIntegrationTest.php:29```

linawolf:
@wouterj  what was exactly was your input? Could you move your test into this repository

wouterj:
Good call, see this branch for a failing test case: https://github.com/phpDocumentor/guides/compare/main...wouterj:guides:issue-305

It also seems to indicate that this does work as expected for emphasis, but not for inline literals.

linawolf:
I cherry picked your tests into #385, some of the issues had already been resolved, I resolved one more issues: templates in inline code caused unwanted newlines.

PR #392, opened 2023-05-21
Introduce Textroles

linawolf:
depends on #390

resolves https://github.com/phpDocumentor/guides/issues/178
references https://github.com/phpDocumentor/guides/issues/305


jaapio:
@linawolf  could you please rebase this PR, that will help me to give you ar proper review. 

I'm really happy to see that you found the energy to get this moving again. Looking forward to reviewing this and seeing how it works!

packages/guides/src/Compiler/Passes/ReferenceResolverPass.php
```diff
@@ -0,0 +1,129 @@
...
+use function sprintf;
+use function str_contains;
+
+/**
+ * Resolves reference tokens in span nodes
+ *
+ * This follows the reStructuredText rules as outlined in:
+ * https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#implicit-hyperlink-targets
+ */
+class ReferenceResolverPass implements CompilerPass
```
jaapio, line 30: I like the movement of this logic to the compiler. That makes a lot of sense. 

What I do see is that you have another case where the NodeTraverser logic doesn't work because you needed the Document. What do you think about adding the `Document` to the parameters you get in `NodeTransformer::enterNode` and `NodeTransformer::leaveNode` I think that would simplify your usecase, as you do not need to have the recursion here. 

Your implementation could be simply

```
class ReferenceResolverNodeTransformer implements NodeTransfomer {

    public function supports(Node $node): bool
    {
        return $node instanceof SpanNode;
    }

     public function enterNode(Node $node, Document $document) {
        return $node;
     }
     
     
     public function leaveNode(Node $node, Document $document) {
              foreach ($child->getTokens() as $token) {
                    if ($token instanceof ReferenceNode) {
                        $this->resolveReference($token, $document);
                    } else {
                        if ($token instanceof DocReferenceNode) {
                            $this->resolveDocReference($token, $document);
                        }
                    }
                }
        
               return $node;
     }
}
```

packages/guides/src/NodeRenderers/Html/SpanNodeRenderer.php
```diff
@@ -14,11 +14,15 @@
 namespace phpDocumentor\Guides\NodeRenderers\Html;
```
jaapio, line 14: This class needs our attention :-) Somehow I think we should transformate the spannode into a proper node structure like all other nodes. That would make it possible to use the logic of renderers and templates for spans. It would remove a lot of weird code from the project. 

Let's discuss this after we merged this PR. Introducing all the templates is already a really great move!

jaapio (changes requested):
This looks really great. I have some comments left, but also some stuff that we can do in a follow up. Thanks for this major improvement!

packages/guides/src/NodeRenderers/Html/SpanNodeRenderer.php
linawolf: It certainly does

packages/guides/src/Compiler/Passes/ReferenceResolverPass.php
linawolf: Ok, I see how you mean this. Another Problem however. We first must run all the CollectLinkTargetsTransformers, then the MetasPass and then the reference resolving. Otherwise the link targets are not present yet at the time running the ReferenceNodeTransformer

linawolf: This means we would have to also introduce a priority system for the Transformers...

linawolf: Have a look at my WIP commit

linawolf:

jaapio (approved):

jaapio:
Thank you very much for this great improvement!

PR #393, opened 2023-05-21
Add configuration file to allow registering custom extensions

wouterj:
This builds on top of #389. You can view the diff from this PR at

This adds a very basic configuration file, to register other extensions. I've used XML as that also seems to be the format used by phpDocumentor.

After this PR, you can create a `guides.xml` file and use the `<extension>` element to register extensions others than Guides and reStructuredText:

```xml
<!-- guides.xml -->
<?xml version="1.0" encoding="UTF-8" ?>
<guides>
    <!-- I've added a "short FQCN" resolver, the FQCN will be "phpDocumentor\Guides\Bootstrap\DependencyInjection\BootstrapExtension" -->
    <extension class="phpDocumentor\Guides\Bootstrap"/>
    <extension class="phpDocumentor\Guides\PhpDocumentor"/>
</guides>
```

The `ContainerFactory` parses this config file and registers the extensions. The `BootstrapExtension` then prepends configuration to register the bootstrap theme. This means guides-cli is now completely decoupled from the theme packages and we can easily add our own custom extensions (e.g. from TYPO3 or Symfony) without doing anything in this repository.

In a next PR, we can introduce a way to set extension configuration through the config file. This will allow us to e.g. configure defaults for the project title or theme, which can then be overridden using CLI options implemented in #368.

As an extra, I've also added support for an `<import>guides.xml.dist</import>` element to import other configuration files.


wouterj:
Btw, I personally think this is as far as I would like to take the standalone CLI project. The important next step will be to see if the changes in my last 3 PRs also makes it easier to integrate this into phpDocumentor. I've tried to always rely on standard Symfony interfaces. In theory, this should make it possible to quite easily connect the DI extensions and configuration in the main phpDocumentor application.
E.g. the phpDocumentor config file would then have a `<guides>` element that works exactly like the `<guides>` root element in the config snippet shown in the description. @jaapio I think it would be good to share our thoughts about this. I can make some time to help you out with the re-integration.

jaapio:
@wouterj could you please rebase this pr. I will review this afterwards, most of the changes have been merged I think?

wouterj:
Oh interesting, I thought git automatically recognized duplicate commits if they had the same hash. Anyway, PR updated and tests fixed now :)

jaapio:
I think we are moving in the right direction however phpDocumentor has it's own configuration structure, we can change this, but just integrating the config might be an issue. But let's see how we can solve this, the same thing applies to the current theme manager implementation, that will be a puzzle to solve in phpDocumentor. I think the way package configuration works in Symfony could help us with this? 

In the case of phpDocumentor, it would be nice if we could make guides an extension of phpDocumentor. I think that the current setup could work as phpDocumentor is a full-blown Symfony application right now. There will be some points that need some customization but that can be done as the service container gives us the options to get the required services. 
Your implementation aligns perfectly with what I was building for extensions in phpDocumentor, but as we are using a cached container this has some extra complexity to kill. But I can live with removing the cache for now I think. 

I suggest merging this PR and seeing how we can do the integration in phpDocumentor. and create some follow up PR's do improve the implementation to match with what we need.

jaapio:
Merged manually, because of the lock file conflicts.

PR #396, opened 2023-05-27
Register Theme Extensions

linawolf:
This way calling
```
ddev exec vendor/bin/guides docs output --theme=bootstrap
```

Works again from the mono-repository. Also this provides a configuration example

### DIFF
--- a/guides.xml
+++ b/guides.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides>
+    <extension class="phpDocumentor\Guides\Bootstrap"/>
+    <extension class="phpDocumentor\Guides\PhpDocumentor"/>
+</guides>


linawolf:
@wouterj Please have a look if it is correct like that

guides.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides>
+    <extensions>
+        <bootstrap>
+            <class>\phpDocumentor\Guides\Bootstrap\DependencyInjection\BootstrapExtension</class>
+        </bootstrap>
+        <phpdocumentor>
+            <class>\phpDocumentor\Guides\PhpDocumentor\DependencyInjection\PhpDocumentorExtension</class>
+        </phpdocumentor>
+    </extensions>
```
wouterj, line 10: This can be simplified as:
```suggestion
    <extension class="phpDocumentor\Guides\Bootstrap"/>
    <extension class="phpDocumentor\Guides\PhpDocumentor"/>
```

The [`fixXmlConfig()` call](https://github.com/phpDocumentor/guides/blob/main/packages/guides-cli/src/Config/Configuration.php#L22) tells the config component that any `extension` element should be interpreted as an item in the `extensions` array option. For a standard key-value pair, it is common to use an attribute in XML (although both attribute and element are supported by the config component).

On top of this, I've added some [logic to the `ContainerFactory`](https://github.com/phpDocumentor/guides/blob/main/packages/guides-cli/src/DependencyInjection/ContainerFactory.php#L97-L114) that allows abbreviating the full class name to a shorter version. `Some\Namespace\Xxx` will be transformed to `Some\Namespace\Xxx\DependencyInjection\XxxExtension`.

PR #397, opened 2023-05-28
Remove phpdoc theme as it is part of phpdoc core

jaapio:
The phpDocumentor template is part of the phpDocumentor core, as it depends on the css and classes defined by phpDoc there is no need to have it here.

The extension has been removed as it is not used anymore.


linawolf (approved):

linawolf:
You might want to mention in this PR where the extension was moved?

PR #398, opened 2023-05-28
Introduce Document Parts and Special Menus

linawolf:
A document part can be used to specify content which should appear in a certain area of the rendered document, for example the footer or a menu. By setting an option on all child nodes rendering of this part can be handled differently by the theme.


packages/guides/src/Nodes/AbstractNode.php
```diff
@@ -89,6 +89,11 @@ public function withOptions(array $options): Node
         return $result;
     }
 
+    public function addOption(string $identifier, int|bool|string|null $option): void
```
jaapio, line 92: I think we should keep the `withOptions` method and not introduce a new method to set options. The immutable behavior of this class was intended to prevent users from changing the nodes by accident. And make it easier to see when the cache should be invalidated.

packages/guides/src/Compiler/NodeTransformers/MenuNodeTransformer.php
```diff
@@ -38,6 +38,10 @@ public function enterNode(Node $node, DocumentNode $documentNode): Node
             }
 
             foreach ($this->buildFromDocumentEntry($metaEntry, 1, $node) as $entry) {
+                if ($entry->getUrl() === $documentNode->getFilePath()) {
```
jaapio, line 41: should this not be view/template behavior? all nodes will become active like this?

jaapio (commented):
I found it interesting to see that you changed some of the core parts of the guides project, but kept the reason for change outside the core. 

Would it harm to introduce the documentblocks in the core? it seems to be a good solution for me to create site elements. I'm not sure yet if this should be part of a theme, because people might want to have this feature, but not use bootstrap as a theme? 

It's an interesting concept for sure. Thanks for introducing this.

packages/guides/src/Compiler/NodeTransformers/MenuNodeTransformer.php
linawolf: This sets a flag on a documentEntry if (and only if) the document entry equals the current document. The view can then decide on if and how to display this information

packages/guides/src/Nodes/AbstractNode.php
linawolf: So, we can add and remove nodes in the nodestransformers, but we cannot change them?

jaapio: Add, replace and remove, if you return a new node the old one will be replaced.

packages/guides/src/Compiler/NodeTransformers/MenuNodeTransformer.php
jaapio: Oh, I think I was wrong here, the menu nodes are not shared between pages. This only applies to metas.

linawolf:

PR #399, opened 2023-05-29
Make Inline Markup Tokens Inherit from Node

linawolf:


packages/guides/src/Nodes/InlineNode.php
```diff
@@ -0,0 +1,36 @@
...
+ * @link https://phpdoc.org
+ */
+
+namespace phpDocumentor\Guides\Nodes;
+
+use phpDocumentor\Guides\Nodes\InlineToken\InlineMarkupToken;
+use phpDocumentor\Guides\Nodes\InlineToken\PlainTextToken;
+
+/** @extends CompoundNode<InlineMarkupToken> */
+class InlineNode extends CompoundNode
```
jaapio, line 20: Should we keep this? Or could we remove it since the new nodes can be added directly to their parent? Like

Paragraph
  InlineNode
     LiteralNode
     HyperLinkNode

Could be simplified to:
Paragraph
     LiteralNode
     HyperLinkNode

packages/guides/src/Nodes/InlineToken/GenericTextRoleToken.php
```diff
@@ -4,7 +4,7 @@
 
 namespace phpDocumentor\Guides\Nodes\InlineToken;
 
-class GenericTextRoleToken extends ValueToken
+class GenericTextRoleToken extends InlineMarkupToken
```
jaapio, line 7: Suggestion: drop the name Token as these are nodes now?

jaapio:
I didn't do a full review, as I assume you are still working on this, but I had some input that I wanted to share with you :-)

packages/guides/src/Nodes/InlineToken/GenericTextRoleToken.php
linawolf: Done

packages/guides/src/Nodes/InlineNode.php
linawolf: The rendering of the InlineNode takes care of  for example not applying spaces etc and it is resused in all Places where Text with Markup is used. The Paragraph adds a <p> tag around a text. So they are different things in my eyes.

jaapio (approved):

tests/Functional/tests/link-span/link-span.html
```diff
@@ -1,3 +1,2 @@
-<p>
-    <a href="http://google.com">Go to <strong>Google</strong></a>
-</p>
+SKIP No subparsing of links
+<p><a href="http://google.com">Go to <strong>Google</strong></a></p>
```
wouterj, line 2: As far as I know, this is also not supported by the rst spec. Should we remove this test?

PR #400, opened 2023-05-29
Fix twig table html template so that it renders nbsp.

sarahmccarthy123:
Also added nbsp to empty table cells in tests


linawolf (approved):

linawolf:
@sarahmccarthy123 Thanks a lot for you contribution! This is bringing us one more step forward!

packages/guides/resources/template/html/body/table.html.twig
```diff
@@ -18,10 +18,12 @@
...
                 {% for column in tableRow.columns %}
-                    <td{% if column.colSpan > 1 %} colspan="{{ column.colSpan }}"{% endif %}{% if column.rowSpan > 1 %} rowspan="{{ column.rowSpan }}"{% endif %}>
+                <td{% if column.colSpan > 1 %} colspan="{{ column.colSpan }}"{% endif %}{% if column.rowSpan > 1 %} rowspan="{{ column.rowSpan }}"{% endif %}>
+                    {%- if column.children is not empty %}
                         {%- for child in column.children -%}
-                            {{- renderNode(child) -}}
-                        {% endfor %}</td>
+                                {{- renderNode(child) -}}
+                        {% endfor %}
+                    {%- else %}&nbsp;{% endif %}</td>
```
jaapio, line 26: This can be simplified by the for else construction supported by twig:

```
    {%- for child in column.children -%}
        {{- renderNode(child) -}}
    {% else %}
         &nbsp;
    {% endfor %}
```
This is not a requirement to be changed before merging this, but just so you know this construct exists.

sarahmccarthy123: Thanks for the tip. I've changed my code accordingly

PR #403, opened 2023-05-30
Add compiler pass for textroles

jaapio:
Allow domain specification in text roles by tags.

This compilerpass for symfony DI allows use to define textroles to be injected in a domain. 

The example below will inject the `AbbreviationTextRole` in the domain `std`, multiple tags can be defined duplicating the `tag` definition. This will inject the role into multiple domains. Without the domain specification a textrole will be added to the default set of textroles.

```
    $container->services()
        ->set(AbbreviationTextRole::class)
        ->tag('phpdoc.guides.parser.rst.text_role', ['domain' => 'std')
```


linawolf (approved):

linawolf:
thanks a lot for providing this! It will come in handy for the TYPO3 theme as well

PR #405, opened 2023-06-01
Allow theme inheritance

wouterj:
@linawolf discovered that the recent theme changes did no longer allow to extend other themes. For instance, the TYPO3 docs could extend the existing bootstrap theme and the Symfony docs can extend the RTD theme.

This PR introduces theme inheritance, which I think is the most modular way of allowing this (you no longer have to hardcode paths to vendor templates). Meanwhile, I also allowed a single theme to pass multiple paths.

The config now allows a couple variants:

```php
$container->prependExtensionConfig('guides', [
    'themes' => [
        // current behavior: a theme with a single path. Result will be: theme path + default paths
        'bootstrap' => dirname(__DIR__, 2) . '/resources/template',

        // same, but with multiple paths
        'bootstrap' => [dirname(__DIR__, 2) . '/resources/template', ...],

        // extend from other templates. result will be: theme paths + parent paths + default paths
        'bootstrap' => [
            'extends' => 'parent-theme',
            'templates' => [dirname(__DIR__, 2) . '/resources/template'],
        ],
    ],
]);

```


jaapio:
We use a chainloader in phpDocumentor, this allows us to overwrite specific files and even extend files using the prefix functionally of twig. `guides::` is the prefix for the base templates of this project for example allowing you to wrap the base template in another template. 

It might be worth it to have a look at that. 
https://twig.symfony.com/doc/2.x/api.html#loaders

linawolf:
@jaapio we already use the chainloader. wouter only introduces a new way to register the template paths with our application

linawolf (approved):

PR #409, opened 2023-06-02
Add shadow tree for compiler purposes

jaapio:
- [x] add method to add new child in the tree
- [x] add method to replace node in the tree
- [x] add method to remove a node from the three
- [ ] add get neighbor method that returns all children of the same parent
- [x] update document traverser to use this stucture
- [x] update nodetransformers to use this structure


jaapio:
@linawolf I created a first draft of the helper tree we talked about today. This should give us more flexibility to move around in the tree during the compiling stage without the need to create the bi-directional node structure in the actual nodes.

linawolf:
@jaapio thanks, this could be a good idea!

linawolf (approved):
Very good! I believe we can merge it

PR #412, opened 2023-06-04
Simple table: Do not check gap after last column

wouterj:
In simple tables, the gap was also checked after the last column. Due to overflow reasons, this meant we were checking the first character of the column instead of the gap after it - always resulting in an error log.

See for example the test output for docs-builder: https://github.com/wouterj/symfony-docs-guides/actions/runs/5168781751/jobs/9310597134#step:5:76


linawolf:
Thanks for this change! Could you maybe supply a test for it? The integration tests now make sure that there have been no warnings or errors

wouterj:
All simple table functional tests were reproducing this issue, so I've added the same "enforce no warning/error logs" treatment to the functional test suite as well.

I've also removed a test that was an exact duplicate and I found an important but somewhat unrelated bug, where all compiler passes were registered twice (I didn't know Symfony automatically detected extensions implementing `CompilerPassInterface` as being compiler passes).

linawolf (commented):
I love this! Thanks you!

jaapio (approved):

PR #418, opened 2023-06-10
Fix inline escaping mechanism

wouterj:
After upgrading to the latest guides, I discovered some issues with escaping (probably caused by #399).

This PR updates the inline lexer to follow the spec: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#escaping-mechanism I've verified the updated tests using Sphinx, and the new changes are equal to what Sphinx provides (we were testing on off-spec behavior as far as I can see)

Finally, I had to update the text roles a bit to conform to this rule in the spec:

> In literal context ([literal blocks](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks) and [inline literals](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-literals), content of the [code](https://docutils.sourceforge.io/docs/ref/rst/directives.html#code), [math](https://docutils.sourceforge.io/docs/ref/rst/directives.html#math), and [raw](https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw) directives, content of the ["raw" role](https://docutils.sourceforge.io/docs/ref/rst/roles.html#raw) and [custom roles](https://docutils.sourceforge.io/docs/ref/rst/directives.html#role) based on it), reStructuredText markup characters lose their semantics so there is no reason to escape them.

I pondered between a couple directions on how to solve this:
(A) Introduce an `AbstractTextRole` + `processContent()` method that removes the backslash escapes, which can be overridden by literal text roles
(B) Introduce a `LiteralTextRole` "marker interface" and use this in `TextRoleRule` to determine if we should pass the raw content
(C) The current solution, passing both raw and cleaned up content.

I went for (C) as: Using (A) is not straightforward given it is not clear what the "content" is outside a text role (e.g. for references, it's only the part in front of `<...>`), I don't like marker interfaces (B) and Docutils is using (C) as well. If you prefer one of the other directions, I'm happy to update the code :)


tests/Functional/tests/table-grid-utf8/table-grid-utf8.html
```diff
@@ -8,11 +8,7 @@
     <tbody>
         <tr>
             <td>x</td>
-            <td>
-                é à
-                <br>
-                é
-            </td>
+            <td>é à é</td>
```
linawolf, line 11: are you sure about this? I believe the old solution was correct as this is how you make multi line content in tables? Did you test this in sphinx?

linawolf:
I like the solution of passing both the raw and processed content. 

What happens when you have a raw textrole with an escaped backtick inside? like for example 
```
:code:`some\`thing` 
```
did you test that in sphinx?

tests/Functional/tests/table-grid-utf8/table-grid-utf8.html
wouterj: Yes, I verified all changed test output with Sphinx. Seems like multi line tables are created using the line block syntax (https://stackoverflow.com/a/13540865/1149495), but I guess this magical syntax is not yet supported by this parser.

wouterj:
> What happens when you have a raw textrole with an escaped backtick inside? like for example

Good one! I've added it to the test and fixed the behavior to match Sphinx's.

linawolf (approved):
I think this is a good solution, thanks a lot

linawolf:
@jaapio is solution C (See above) also your prefered one?

jaapio:
I think this is a good solution 👍

PR #419, opened 2023-06-11
Fix no-eager literals test

wouterj:


linawolf (approved):
Thanks a lot, I was searching for a solution like that and couldn't find one!

PR #424, opened 2023-06-20
Remove role from raw content with domain

jaapio:
Added a unittest for the TextRoleRule to cover the use cases including the broken behavior that added the role when a role was defined with a domain.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/InlineRules/TextRoleRule.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/InlineRules/TextRoleRule.php
@@ -34,7 +34,6 @@ class TextRoleRule extends AbstractInlineRule
 
         $initialPosition = $lexer->token?->position;
         $lexer->moveNext();
-
         while ($lexer->token !== null) {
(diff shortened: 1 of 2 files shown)


packages/guides-restructured-text/src/RestructuredText/Parser/Productions/InlineRules/TextRoleRule.php
```diff
@@ -34,15 +34,14 @@ public function apply(ParserContext $parserContext, InlineLexer $lexer): InlineN
...
-
         while ($lexer->token !== null) {
             $token = $lexer->token;
             switch ($token->type) {
                 case $token->type === InlineLexer::COLON && $inText === false:
                     if ($role !== null) {
                         $domain = $role;
                         $role = $part;
-                        $part = '';
+                        $rawPart = $part = '';
```
jaapio, line 44: This was the missing line

linawolf: wow

linawolf (approved):
Thanks for catching this and for adding the tests!

PR #430, opened 2023-06-27
Fix url resolving

jaapio:
Urls should start with a / like the toc urls do. This allows people to use the base url metadata in document header to overwrite the actual location of the generated documents.


linawolf:
It might be a good idea to offer both relative and absolute urls and make them configurable. In TYPO3 we have all manuals at the same domain...

linawolf (approved):

wouterj:
Good idea! (Both this PR and Lina's suggestion)

I'm working on refactoring url resolving a bit. There are some parsing issues in the current impl, and to fix this nicely I want to introduce a new resolving abstraction in the render phase. I should have a POC ready before Monday, so maybe we can discuss it a bit then.

PR #439, opened 2023-07-03
Create files with host user ids

greg0ire:
When running docker with the php image, any command is run as root, unless. It is possible to specify another user to run as with the `--user` option though.
Note that creating the user and group on the docker image does not seem to be required for this to work, so let us just not do it. This spares us from creating a custom entrypoint file.


linawolf (approved):

linawolf:
Thanks a lot! this saves me some headache!

PR #440, opened 2023-07-03
Specify cache directory for Psalm

greg0ire:
Before 90fb333, the Psalm cache directory was presumably created in the home directory of the root user.

After that commit, Psalm attempts to create that directory in /, and that fails because of permission issues.

Specifying a cache directory:
- fixes that particular issue;
- more importantly, allows the cache directory to be reused.

Note that `--update-baseline` does not seem to benefit from the cache, so this is really beneficial only when running `vendor/bin/psalm` (no option).

This fixes `make psalm`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 /.build
 .phpunit.cache
 .deptrac.cache
+/.psalm
 /**/temp
 /output/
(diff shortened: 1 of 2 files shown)


linawolf (approved):

linawolf:
Thanks for this fix

Issue #297, opened 2023-04-15
Add progress bar to cli


jaapio:
When executing the cli app, there is no output shown, introduce a progress bar to show some output to the user.

greg0ire:
I'll look into this issue :slightly_smiling_face:

PR #441, opened 2023-07-03
Add progress bar to the run command

greg0ire:
This adds a progress bar based on how many documents are being processed.

This is not always possible depending on the capabilities of the current `OutputInterface` interface implementation, which it is conditionally created.

For instance, when running tests, you get a `BufferedOutput`, which does not have the section capability.

Closes https://github.com/phpDocumentor/guides/issues/297

### DIFF
--- a/packages/guides/src/Handlers/RenderCommand.php
+++ b/packages/guides/src/Handlers/RenderCommand.php
@@ -10,10 +10,10 @@ use phpDocumentor\Guides\Nodes\ProjectNode;
 
 final class RenderCommand
 {
-    /** @param DocumentNode[] $documents */
+    /** @param iterable<DocumentNode> $documents */
     public function __construct(
         private readonly string $outputFormat,
-        private readonly array $documents,
+        private readonly iterable $documents,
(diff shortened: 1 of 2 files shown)


packages/guides/src/Handlers/RenderCommand.php
```diff
@@ -10,10 +10,10 @@
 
 final class RenderCommand
 {
-    /** @param DocumentNode[] $documents */
+    /** @param iterable<DocumentNode> $documents */
```
greg0ire, line 13: `iterate` returns a generator, but looking at the rest of a project, the fact that `getDocuments()` is an array is never used.

linawolf: I think it should be fine then

linawolf (approved):
Looks good to me

greg0ire:
You're right, I'll work on that 👍

greg0ire:
@jaapio done :slightly_smiling_face:

jaapio:
Thanks you rock!

jaapio (approved):

PR #443, opened 2023-07-03
Create a Document Tree and the Menus from it

linawolf:
Menu entries are nodes now
Document entries are nodes now. They have been moved from the Metas to the project nodes

Menus are resolved during Transformation. 
This will enable us to deal with multi-level menus in the sphinx style in the future


jaapio:
@linawolf  could you please add a description to this PR to explain what you changed?

linawolf:
@jaapio I tried to add more information. Could you try to help me what you would expect in the description?

jaapio:
- What is the reason the entries are now nodes? 
- How do the changes you did for the resolving of menus relate to the work Wouter did? Is it just the building of the menu tree that you are doing in the compiler right now?

linawolf:
In the compiler I am Building the general document tree. In a follow up we can integrate the sphinx functionality that Document Entries can be moved in the Document Tree by their definition in doctrees. Therefore this tree does not rely on the file structure anymore.

Then for each toctree I am building a MenuTree which only contains those MenuEntries that the menu should display with its settings. 

Each menu entry has the information about which file and anchor on that file it should reference. Therefore we can then integrate the url resolution implemented by wouter

packages/guides/src/Nodes/DocumentTree/DocumentEntryNode.php
```diff
@@ -0,0 +1,65 @@
...
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Nodes\DocumentTree;
+
+use phpDocumentor\Guides\Nodes\AbstractNode;
+use phpDocumentor\Guides\Nodes\DocumentNode;
+use phpDocumentor\Guides\Nodes\TitleNode;
+
+/** @extends AbstractNode<DocumentNode> */
+class DocumentEntryNode extends AbstractNode
```
jaapio, line 12: Please mark this class as final. People should not extend it.

packages/guides/src/Nodes/DocumentTree/SectionEntryNode.php
```diff
@@ -2,13 +2,13 @@
 
 declare(strict_types=1);
 
-namespace phpDocumentor\Guides\Meta;
+namespace phpDocumentor\Guides\Nodes\DocumentTree;
 
 use phpDocumentor\Guides\Nodes\TitleNode;
 
-class SectionEntry implements ChildEntry
+class SectionEntryNode
```
jaapio, line 9: This class has the suffix `Node` but it's not a node?

jaapio (commented):
I think there are some optimizations possible in the code as suggested. 

I didn't find any big issues here. Thanks for these improvements!

packages/guides/src/Compiler/NodeTransformers/SectionEntryRegistrationTransformer.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function supports(Node $node): bool
+    {
+        return $node instanceof SectionNode;
+    }
+
+    public function getPriority(): int
+    {
+        // After DocumentEntryRegistrationTransformer
```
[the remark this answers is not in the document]
linawolf: the DocumentEntryRegistrationTransformer still exists

packages/guides/src/Nodes/DocumentTree/SectionEntryNode.php
linawolf: I think I changed that in the next Pull Request where I noticed it when I could not transform it. But changed it now

linawolf:

jaapio (approved):

jaapio:
This is great! thanks!

PR #446, opened 2023-07-03
Introduce rst-class directive as alias for class directive

linawolf:
just like sphinx is doing


packages/guides-restructured-text/src/RestructuredText/Directives/RstClassDirective.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * When the default domain contains a class directive, this directive will be shadowed. Therefore, Sphinx re-exports it as rst-class.
+ *
+ * See https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#rstclass
+ */
+class RstClassDirective extends ClassDirective
+{
+    public function getName(): string
+    {
+        return 'rst-class';
```
jaapio, line 29: There is no need for an extra class, as directives support the getAliases() method. You can simply add an alias there.

linawolf: yeah right, forgot about that one, thanks!

jaapio (approved):

PR #449, opened 2023-07-03
Be stricter about what happens in tests

greg0ire:
This ensures we cannot have a PHP notice or warning in the test suite. Also, we display details when that happens, for easier debugging.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -8,6 +8,11 @@
         beStrictAboutTestsThatDoNotTestAnything="true"
         beStrictAboutOutputDuringTests="true"
         beStrictAboutChangesToGlobalState="true"
+        failOnNotice="true"
+        failOnWarning="true"
+        displayDetailsOnTestsThatTriggerNotices="true"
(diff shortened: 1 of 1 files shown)


greg0ire:
Note: there is no `failOnError` attribute. When using `trigger_error('error', E_USER_ERROR)`, the test suite does not fail, even with this PR. But since in recent versions of PHP, errors are throwable, maybe that explains why this isn't caught? Anyway, what we really want to catch here are actual notices, warnings and errors, not stuff artificially triggered with `trigger_error`.

jaapio:
I think it's good to have this in. I didn't see any errors does that mean that the library is already clean?

greg0ire:
> I think it's good to have this in. I didn't see any errors does that mean that the library is already clean?

It is already clean indeed, but there is an untested use case for which I'm currently getting a warning, I should send a  PR to fix it today.

linawolf (approved):
I think this is good to be merged once the tests are green

jaapio (approved):

PR #451, opened 2023-07-04
Avoid PHP Warning when parsing empty list item

greg0ire:
Relates to https://github.com/doctrine/orm/pull/10812


jaapio (approved):
Nice, thanks for this addition.

PR #453, opened 2023-07-04
Upgrade to doctrine/lexer 3

greg0ire:
This should make it easier to install the project on projects with recent dependencies.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "bf01f5a5f20fea2d3b08cdb940bf5576",
+    "content-hash": "31cc06f59c0e04b42a580db2e7ab4433",
     "packages": [
         {
             "name": "dflydev/dot-access-data",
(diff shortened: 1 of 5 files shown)


greg0ire:
I would like to have a CI job on all Doctrine projects with the guides-cli running in it, so it's possible to forbid contributors to contribute invalid docs (based on the warnings output to `stderr`). This is one step towards doing that, another step will be to upgrade to Symfony 6

composer.lock
```diff
@@ -1120,7 +1072,7 @@
         },
         {
             "name": "phpdocumentor/guides-cli",
-            "version": "dev-main",
+            "version": "dev-lexer-3",
```
greg0ire, line 1075: I don't know how to avoid this :thinking:

greg0ire:
@wouterj I restored support for `doctrine/lexer` 2 so as not to cause issues similar to the one you are describing here.

packages/guides-restructured-text/composer.json
```diff
@@ -22,7 +22,7 @@
     "minimum-stability": "stable",
     "require": {
         "php": "^8.1",
-        "doctrine/lexer": "^2.1",
+        "doctrine/lexer": "^3.0",
```
wouterj, line 25:
```suggestion
        "doctrine/lexer": "^2.1 || ^3.0",
```

Btw, I think only supporting doctrine/lexer 3 is fine (as far as I can see in https://github.com/phpDocumentor/phpDocumentor , only the guides package is using Doctrine lexer).

wouterj (approved):

packages/guides-restructured-text/composer.json
greg0ire: You're right, I will change that.

greg0ire: @wouterj I dropped `doctrine/lexer` 2 support again.

jaapio:
Thanks for the upgrade!

jaapio (approved):

PR #454, opened 2023-07-04
Address deprecation of spaceless tag

greg0ire:
Since Twig 2.7, the `spaceless` tag is deprecated in favor of the `spaceless` filter, or the `apply` tag with a `spaceless` argument. I picked the latter so as to still benefit from syntax highlighting.

This is the only deprecation I found while running the test suite with `symfony/phpunit-bridge` locally.

### DIFF
--- a/packages/guides/resources/template/html/body/menu/content-menu.html.twig
+++ b/packages/guides/resources/template/html/body/menu/content-menu.html.twig
@@ -1,5 +1,5 @@
-{% spaceless %}
+{% apply spaceless %}
     <div class="contents">
         {% include "body/menu/menu-level.html.twig" %}
     </div>
-{% endspaceless %}
+{% endapply %}
--- a/packages/guides/resources/template/html/body/menu/menu.html.twig
+++ b/packages/guides/resources/template/html/body/menu/menu.html.twig
@@ -1,5 +1,5 @@
-{% spaceless %}
+{% apply spaceless %}
     <div class="menu">
         {% include "body/menu/menu-level.html.twig" %}
     </div>
-{% endspaceless %}
+{% endapply %}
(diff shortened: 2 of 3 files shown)


linawolf (approved):
thanks for catching this!

linawolf:
By running a quick code search there seem to be more then just this one position of spaceless

wouterj:
Where possible, I would suggest to not use spaceless at all and either not care about the resulting HTML or use the [whitespace control characters](https://twig.symfony.com/doc/3.x/templates.html#whitespace-control) where whitespace matters in rendering (e.g. inline or in `pre` elements).

See also this comment: https://github.com/symfony-tools/docs-builder/pull/80#issuecomment-805878833

greg0ire:
@linawolf good catch! Fixed

@linawolf @wouterj looking at occurrences of `spaceless` in the project, I'm not sure I 100% understand what it does. For instance, is it really needed here: https://github.com/phpDocumentor/guides/blob/86aa1e7db978ac7f454e2712046b34fe94a5a90f/packages/guides/resources/template/html/inline/newline.html.twig#L1

I'm starting to wonder if we couldn't remove it in most cases.

greg0ire:
We @wouterj I just talked to @linawolf , most tags are really necessary for inline tags, because for instance if you have `**bold text**.` followed by a dot, since most developers add a newline at EOF, there will be a newline at EOF in the template for the bold html inline tag. The end result will be this:
```html
<p>
<strong>bold text</strong>
.
</p>
```

This means an unwanted space will appear between the text and the dot.

linawolf:
We do need spaceless tags in inline node rendering though. Most IDEs will automatically add an empty line at the end of each file. So twig inserts a newline after each textrole. HTML then interprets this newline as space even if there should be no space between the end of the textrole (for example bolt text) and the next char (for example a dot or brace).

greg0ire:
Anyway, I think this is orthogonal to that PR

linawolf (approved):

PR #459, opened 2023-07-05
Fail on log

greg0ire:
Currently, when all is well, nothing should be logged at all. We can leverage this to use the CLI tool as a linter, and create CI jobs with it.
Here, a new option called `--fail-on-log` is introduced, and it will use a failure exit code as soon as any log is emitted.
This is based on the assumption that the logger is never and will never be called for anything that should not result in a change in the documentation being processed.

### DIFF
--- a/packages/guides-cli/src/Logger/SpyProcessor.php
+++ b/packages/guides-cli/src/Logger/SpyProcessor.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Cli\Logger;
+
+use Monolog\Processor\ProcessorInterface;
+
+/**
+ * This decorator has an extra method to check whether anything was logged
+ *
+ * @internal
+ */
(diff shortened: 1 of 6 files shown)


packages/guides-cli/src/Logger/SpyProcessor.php
```diff
@@ -0,0 +1,30 @@
...
+namespace phpDocumentor\Guides\Cli\Logger;
+
+use Monolog\Processor\ProcessorInterface;
+
+/**
+ * This decorator has an extra method to check whether anything was logged
+ *
+ * @internal
+ */
+final class SpyProcessor implements ProcessorInterface
```
greg0ire, line 14: I first decorated `LoggerInterface` to do this, and then realized that `Run` depends on `Logger`, which has a much bigger interface. I then switched to decorating a processor, which feels better because the interface is tiny, and there is no decoration involved.

linawolf:
I love this!!

jaapio (approved):

PR #460, opened 2023-07-05
Try fixed workflow

greg0ire:
Doing so involves switching from the `dev-main@dev` version constraint to
`self.version`, because now we run Composer with different options, and it
behaves errors out for some of them if we stick with `dev-main@dev`


composer.lock
```diff
@@ -1026,7 +1026,7 @@
         },
         {
             "name": "phpdocumentor/guides",
-            "version": "dev-main",
+            "version": "dev-fixed-workflow",
```
greg0ire, line 1029: Argh! This is bad…

jaapio (approved):

PR #455, opened 2023-07-04
Introduce Transformer for Content Menus 

linawolf:
depends on #443

PR #461, opened 2023-07-05
Enable Multi level Menus

linawolf:
To avoid circular relations we need to mark the root document.

depends on #455, which depends on https://github.com/phpDocumentor/guides/pull/443


jaapio:
@linawolf  please rebase this one, so we can continue with a review :-)

linawolf:
@jaapio rebased, please review

packages/guides/src/Compiler/NodeTransformers/TocNodeWithDocumentEntryTransformer.php
```diff
@@ -96,42 +126,59 @@ public function getPriority(): int
...
+        if ($expectedFile === '/' . $actualFile) {
             return true;
         }
 
-        return $this->isGlob($glob, $documentEntry->getFile(), $currentPath, $file, '/');
+        return self::isGlob($glob, $actualFile, $currentFile, $expectedFile, '/');
     }
 
-    private function isEqualRelativePath(DocumentEntryNode $documentEntry, string $file, string $currentPath, bool $glob): bool
+    public static function isEqualRelativePath(string $actualFile, string $expectedFile, string $currentFile, bool $glob): bool
```
jaapio, line 142: Please keep these methods private, as they are not used, and should not be used on other locations.

packages/guides/src/Compiler/NodeTransformers/TocNodeWithDocumentEntryTransformer.php
```diff
@@ -96,42 +126,59 @@ public function getPriority(): int
         return 4500;
     }
 
-    private function isEqualAbsolutePath(DocumentEntryNode $documentEntry, string $file, string $currentPath, bool $glob): bool
+    public static function isEqualAbsolutePath(string $actualFile, string $expectedFile, string $currentFile, bool $glob): bool
```
jaapio, line 129: Please keep these methods private, as they are not used, and should not be used on other locations.

packages/guides/src/Compiler/NodeTransformers/TocNodeWithDocumentEntryTransformer.php
```diff
@@ -45,7 +52,7 @@ public function leaveNode(Node $node, CompilerContext $compilerContext): Node|nu
 
         foreach ($files as $file) {
             foreach ($documentEntries as $documentEntry) {
-                if (!$this->isEqualAbsolutePath($documentEntry, $file, $currentPath, $glob) && !$this->isEqualRelativePath($documentEntry, $file, $currentPath, $glob)) {
+                if (!$this->isEqualAbsolutePath($documentEntry->getFile(), $file, $currentPath, $glob) && !$this->isEqualRelativePath($documentEntry->getFile(), $file, $currentPath, $glob)) {
```
jaapio, line 55: isEqualAbsolutePath is static so is isEqualRelativePath. Calls should be `self::` not using `$this`

packages/guides/src/Handlers/ParseDirectoryHandler.php
```diff
@@ -29,7 +29,7 @@ public function handle(ParseDirectoryCommand $command): array
         $currentDirectory = $command->getDirectory();
         $extension = $command->getInputFormat();
 
-        $this->guardThatAnIndexFileExists(
+        $indexName = $this->guardThatAnIndexFileExists(
```
jaapio, line 32:
```suggestion
        $indexName = $this->getDirectoryIndexFile(
```

jaapio (changes requested):
I think this is good for now, some minor things that need to be addressed before we can merge this. 

---
But as I said on Slack. I think we should have a closer look at what we want to store in the nodes. It feels like they are getting larger and larger every feature we add. But most of the new meta like information that is now stored in the documentnode should not be directly used by the renderer. Which means for me that it should not be in the nodes. It is more like our internal administration that nobody should ever look at. 

During rendering the renderer should basically just look at nodes and child nodes. All properties exposed by the nodes should add value to the rendering. And I think the DocumentEntryNode doesn't do that. We want people to use the menu nodes an not trying to create there own menu tree in the templates to render a menu. 

I think we can address this later.

linawolf:

linawolf:
@jaapio I applied all suggested changes, please have another look

jaapio (approved):

PR #470, opened 2023-07-05
Move test for unknown directives to Integration Tests

linawolf:
Here we can test for warnings. There are no warnings, created an issue for that


jaapio:
Thanks!

PR #475, opened 2023-07-06
Add guides-code extension

greg0ire:
The original code is in https://github.com/wouterj/symfony-docs-guides


packages/guides-cli/src/Command/Run.php
```diff
@@ -47,6 +48,7 @@ public function __construct(
         private readonly Logger $logger,
         private readonly ThemeManager $themeManager,
         private readonly SettingsManager $settingsManager,
+        private readonly Highlighter $highlighter,
```
greg0ire, line 51: Please disregard this change, it's just to figure out what I have.

linawolf:
@greg0ire are you still working on this?

greg0ire:
@linawolf no because it needs https://github.com/phpDocumentor/guides/pull/482

linawolf:
@greg0ire https://github.com/phpDocumentor/guides/pull/482 is merged

greg0ire:
Thanks Lina, I'll try to take another look at it soon, see what more needs to be done (IIRC there was a lot more to be done).

greg0ire:
I pushed a version where the highlighter is now configurable thanks to @wouterj 's PR. It's plugged to a Twig extension that exposes a `highlight` filter. I'm not sure exactly what the next steps are. I suppose what's left is to create a template for the `code-block` role, and have it call `highlight`? I'm also not sure how to include the required css files for this to work.

Maybe we should merge as is, and then @wouterj tries removing the `Highlighter` in https://github.com/wouterj/symfony-docs-guides and using this one instead, by requiring the new extension? And then check that everything works like before? And then somebody, not necessarily me, does the plumbing?

linawolf:
@greg0ire your extension can override the template which would also enable us to test the output.

Regarding the css could we include them via https://cdn.jsdelivr.net/ or some such like we do with the css and js for bootstrap in the bootstrap package? Systems who use a package manager might want to build their own minified css anyway.

linawolf:
@jaapio should we get this merged?

jaapio:
I think we can merge this, however there seem to be missing parts, as there is nothing done yet regarding styles? 
This can be done in a separate PR, but I don't expect this to work right now.

jaapio:
Thanks for this contribution @greg0ire, I'm looking forward to see this work. Who ever has time to finish the work to be done.

greg0ire:
You're welcome :) I think it's unlikely I will be the one finishing this, I have a lot to do with Doctrine, although I'm willing to discuss the missing parts. This is not usable by end users right now, but I think it could be used by @wouterj 's project as described in https://github.com/phpDocumentor/guides/pull/475#issuecomment-1722220772 … that might be a good next step.

wouterj:
I've tried this out in my project and made a couple changes (adding some config options) in #634 

I'm not sure if we should ship styles in the extension to be honest, as we're also not using the Twig filter in any of the build-in themes. When a theme uses the filter, it can choose any Highlight.js theme (or create a custom one).

PR #480, opened 2023-07-06
Reference resolving

jaapio:


packages/guides/src/ReferenceResolvers/PreRenderReferenceResolver.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace phpDocumentor\Guides\ReferenceResolvers;
+
+use phpDocumentor\Guides\NodeRenderers\PreRenderers\PreNodeRenderer;
+use phpDocumentor\Guides\Nodes\Inline\LinkInlineNode;
+use phpDocumentor\Guides\Nodes\Node;
+use phpDocumentor\Guides\RenderContext;
+use Webmozart\Assert\Assert;
+
+final class PreRenderReferenceResolver implements PreNodeRenderer
```
wouterj, line 13: I think this should be called a `ReferenceResolverPreRenderer` instead?

PR #482, opened 2023-07-07
Add support for extension configuration

wouterj:


greg0ire:
I think the XML snippet in the PR description could be used in https://github.com/phpDocumentor/guides/blob/main/docs/configuration.rst

linawolf:
@wouterj could we integrate intersphinx configuration (https://github.com/phpDocumentor/guides/pull/494) into this somehow?

linawolf:
@wouterj @greg0ire what is needed to make this work?

greg0ire:
@linawolf I don't understand the question. Are you saying it does not work?

linawolf:
Some tests are still red and it needs to be rebased so we can merge it

greg0ire:
Ok: what needs IMO to be done for this to be mergeable:

- fix the cs issue
- either address or ignore the static analysis issue
- add documentation
- add tests if possible

linawolf:
@wouterj @greg0ire I was trying to add a functional test, however it seems like it has not effect. I don't really understand how this is supposed to work so could someone :tm: support me here?

greg0ire:
This PR allows to build a configurable highlighter, but does not build it itself, that's part of what https://github.com/phpDocumentor/guides/pull/475 is for.

linawolf:
@greg0ire is there anything that we can already test with an integration test? I still don't understand how this works or how to use the configuration. Not even if my example file is named correctly and in the correct place

wouterj:
The second snippet with `<project title="Project Title" version="6.4"/>` is already implemented.

Each extension defines its configuration in the `getConfigTreeBuilder()` method: https://github.com/phpDocumentor/guides/pull/482/files#diff-a98cf5815ab9d23df43257abd002a77c5792f02cff4ef0ee5add504533a9d8cbR36

The Config component parses the XML and finds the nested nodes from the config tree based on this. So for a `<project>` XML element, it'll find the `arrayNode('project')` and the `title="..."` attribute is mapped to the `scalarNode('title')` child.

This is then processed by the Config component (https://github.com/phpDocumentor/guides/pull/482/files#diff-a98cf5815ab9d23df43257abd002a77c5792f02cff4ef0ee5add504533a9d8cbR71-R72) and transformed into one merged associative array like so:

```php
[
  'project' => [
    'title' => '...',
  ]
]
```

linawolf:
@wouterj, I updated the integration test accordingly, unfortunate it does not seem to work

greg0ire:
@linawolf it looks like this is the first time a `guides.xml` file is created in a test. The only place where `guides.xml` is referenced is here:  https://github.com/phpDocumentor/guides/blob/a824e134a68993ebca103c5daa179a1c9fc297cf/packages/guides-cli/bin/guides#L33-L35

but the CLI does not get run when in tests, right? So I don't think the `guides.xml` in this PR will be taken into account at all.

linawolf:
@greg0ire so I am loading the guides.xml in a test now if it is present, however it still does not seem to do anything...

wouterj:
Fyi: I'm taking a look now

wouterj:
Fixed the tests, added some docs, added the XML schema and did some other minor improvements to the project settings flow.

I think this is ready now.

docs/configuration.rst
```diff
@@ -4,12 +4,54 @@
...
+
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="UTF-8" ?>
+    <guides xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd">
+
+        html-theme="bootstrap"
+    >
```
greg0ire, line 50: :thinking: I think these 2 lines have a problem… should they be part of the `extension` tag?

docs/configuration.rst
```diff
@@ -4,12 +4,54 @@
...
+you want to use the Bootstrap HTML theme, use this configuration:
+
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="UTF-8" ?>
+    <guides xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+
+        html-theme="bootstrap"
```
greg0ire, line 49: So this means 2 extensions should be careful not to have conflicting attributes?

wouterj: The `GuidesExtension` is special. Anything in the root that is not `<extension>` or `<import>` will be passed to the `GuidesExtension`, which is also why the `<project>` node is possible.

Other extensions must be configured through their `<extension>` tag, e.g:

```xml
<guides>
  <extension class="SymfonyTools\Symfony"
    default-highlight-language="php"
  />
</guides>
```

greg0ire (approved):

linawolf:
Looks good to me, thanks a lot for these changes

PR #483, opened 2023-07-07
Add docs workflow

FlorianT03:
Adding the documentation.yml to the workflows, but removed some unnecessary jobs (Remove existing composer file, Require phpdocumentor/guides-cli, Configure minimum stability)

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,38 @@
+name: "Documentation"
+
+on:
+  pull_request:
+    branches:
+      - "*.x"
+    paths:
+      - .github/workflows/documentation.yml
+      - docs/**
+  push:
+    branches:
+      - "*.x"
+    paths:
+      - .github/workflows/documentation.yml
+      - docs/**
+
+jobs:
+  validate-with-guides:
+    name: "Validate documentation with phpDocumentor/guides"
+    runs-on: "ubuntu-22.04"
+
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@v3"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "8.2"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v2"
+        with:
+          dependency-versions: "highest"
+
+      - name: "Run guides-cli"
+        run: "vendor/bin/guides -vvv --no-progress docs/en /tmp/test 2>&1 | ( ! grep WARNING )"


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,38 @@
...
+          coverage: "none"
+          php-version: "8.2"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v2"
+        with:
+          dependency-versions: "highest"
+
+      - name: "Run guides-cli"
+        run: "vendor/bin/guides -vvv --no-progress docs/en /tmp/test 2>&1 | ( ! grep WARNING )"
```
greg0ire, line 38: We can now use `--fail-on-log`, it is more simple :tada: 

```suggestion
        run: "vendor/bin/guides -vvv --no-progress docs/en /tmp/test --fail-on-log"
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,38 @@
+name: "Documentation"
+
+on:
+  pull_request:
+    branches:
+      - "*.x"
```
greg0ire, line 6: The naming scheme is not the same as Doctrine, this needs to be changed.

```suggestion
      - "main"
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,38 @@
...
+    paths:
+      - .github/workflows/documentation.yml
+      - docs/**
+  push:
+    branches:
+      - "*.x"
```
greg0ire, line 12:
```suggestion
      - "main"
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,38 @@
+name: "Documentation"
+
+on:
```
greg0ire, line 3:
```suggestion
on: # yamllint disable-line rule:truthy
```

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+    branches:
+      - "*.x"
+    paths:
+      - .github/workflows/documentation.yml
+      - docs/**
```
greg0ire, line 9: Yamllint says you should quote every string with double quotes. that includes these 2 and a few others below as well.

.github/workflows/documentation.yml
FlorianT03: Good to know!

.github/workflows/documentation.yml
FlorianT03: I keep that in mind!

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,38 @@
...
+          coverage: "none"
+          php-version: "8.2"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v2"
+        with:
+          dependency-versions: "highest"
+
+      - name: "Run guides-cli"
+        run: "vendor/bin/guides -vvv --no-progress docs/en /tmp/test --fail-on-log"
```
greg0ire, line 38: The `en` directory exists only in Doctrine.

```suggestion
        run: "vendor/bin/guides -vvv --no-progress docs /tmp/test --fail-on-log"
```

greg0ire:
@linawolf @jaapio can you please approve the workflow run?

.github/workflows/documentation.yml
greg0ire: You need to address this comment in order to fix the build.

jaapio:
Looks like everything is green now. Thanks for your contribution!

PR #492, opened 2023-07-13
Fix performance of InlineLexer by delaying URI scheme checking

wouterj:


tests/Functional/tests/guilabel/guilabel.html
```diff
@@ -1,3 +1,3 @@
 <p>Lorem ipsum dolor sit amet <span class="guilabel">Foo Bar</span>.</p>
-<p>Lorem ipsum dolor sit amet<span class="guilabel">Foo Bar</span>.</p>
+<p>Lorem ipsum dolor sit amet:guilabel:<code>Foo Bar</code>.</p>
```
wouterj, line 2: This is a change in behavior, but it's compatible with Sphinx. It also detects ``amet:guilabel:`something` `` as being 2 separate nodes: `amet:guilabel:` and `` `something` ``.

jaapio:
@linawolf these changes look good to me, but I would like to have your opinion on this as you implemented the guidable behavior.

linawolf (approved):

PR #496, opened 2023-07-16
Introduce menu directive

linawolf:
Toctrees and navigational menus need to add differently in certain aspects:

- TOCS in globmode do not add the current page to the glob. This is to avoid adding the index on the level of the other files in that directory, which are regarded as children of index. TOCS change the document tree, however we want to have a menu on each and every page without having to change the document tree
- The new MENU directive displays a menu (by default of the main level) without subsections (titlesonly always active) always with glob and it DOES NOT change the document tree
- Fixed the bootstrap template for subpage menus, it did not work properly yet.
- get rid of the warnings on rendering our own docs that were related to using toctrees on every page


packages/guides/src/Compiler/NodeTransformers/MenuNodeAddEntryTransformer.php
```diff
@@ -184,4 +172,55 @@ public static function isAbsoluteFile(string $expectedFile): bool
...
+            }
+
+            $documentEntryInToc->setParent($compilerContext->getDocumentNode()->getDocumentEntry());
+            $compilerContext->getDocumentNode()->getDocumentEntry()->addChild($documentEntryInToc);
+        }
+    }
+
+    private function addSubSectionsToMenuEntries(DocumentEntryNode $documentEntry, MenuEntryNode $menuEntry): void
+    {
+        foreach ($documentEntry->getSections() as $section) {
```
jaapio, line 208: Does a document always have a single main section? I think we should just ignore the first level 1 section. Others should be added.

packages/guides/src/Nodes/Menu/NavMenuNode.php
```diff
@@ -0,0 +1,38 @@
...
+ * file that was distributed with this source code.
+ *
+ * @link https://phpdoc.org
+ */
+
+namespace phpDocumentor\Guides\Nodes\Menu;
+
+use function is_scalar;
+
+/** @link https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#table-of-contents */
```
jaapio, line 18: This link seems to be invalid for this class?

packages/guides/src/Compiler/NodeTransformers/MenuNodeAddEntryTransformer.php
```diff
@@ -184,4 +172,55 @@ public static function isAbsoluteFile(string $expectedFile): bool
...
+        string $currentPath,
+    ): void {
+        foreach ($documentEntriesInTree as $documentEntryInToc) {
+            if ($documentEntryInToc->isRoot() || $currentPath === $documentEntryInToc->getFile()) {
+                // The root page may not be attached to any other
+                continue;
+            }
+
+            if ($documentEntryInToc->getParent() !== null && $documentEntryInToc->getParent() !== $compilerContext->getDocumentNode()->getDocumentEntry()) {
+                $this->logger->warning(sprintf(
```
jaapio, line 189: Why is this a warning? Is this an issue somehow that should be fixed? If so, please extend the log message with extra information so the users are aware what is wrong.

packages/guides/src/Compiler/NodeTransformers/MenuNodeAddEntryTransformer.php
linawolf: There could be more then one section if two headings of level 1 are used. It is not common or recommended but can happen

packages/guides/src/Compiler/NodeTransformers/MenuNodeAddEntryTransformer.php
linawolf: The warning was not added here but moved in refactoring. great idea anyway

packages/guides/src/Nodes/Menu/NavMenuNode.php
linawolf: true, removed it, was copy and paste

PR #502, opened 2023-07-22
Orphaned pages do not enforce uniqe title

linawolf:


linawolf:
@greg0ire this is the orphan tag that I told you about. With this change you can just add the `:orphan:` tag at the beginning of those files that are not displayed in the doctree and need no title.

greg0ire (approved):

PR #503, opened 2023-07-23
[BUGFIX] Require symfony/http-client directly in the guides

linawolf:
Otherwise it will be missing when installing the guides via composer


jaapio:
I think this requirement should be part of guides-cli. As the library should be open for other clients IMO. 

So maybe it should just require psr/http-client? 

https://packagist.org/providers/psr/http-client-implementation

linawolf:
As long as the symfony http-client is used in the guides it should be required here: https://github.com/phpDocumentor/guides/blob/main/packages/guides/resources/config/guides.php#L49

If we move the dependency to guides-cli, I don't mind. But for now guides is the place where it is used

jaapio:
Let's fix this in a follow up cleanup

jaapio (approved):
Feel free to merge this

PR #504, opened 2023-07-23
[BUGFIX] Prevent Duplicates in toctree and menu

linawolf:
A page that has already been explicitely added to a toctree does not get added by the glob directive once more. As having multiple occurences of the same page within one toctree or menu does not make sense in general, I applied array_unique to the menu_items of one level of the toctree or menu.

@wouterj, @greg0ire is this behaviour also wanted in the doctrine and symfony docs?


greg0ire:
AFAIK that's not something Doctrine would care about: it does not look like we use globs at all.

jaapio (approved):

PR #514, opened 2023-07-30
[FEATURE] Enable configurable custom and default text roles

linawolf:
See https://docutils.sourceforge.io/docs/ref/rst/directives.html#role

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Directives/DefaultRoleDirective.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Directives/DefaultRoleDirective.php
@@ -16,7 +16,7 @@ namespace phpDocumentor\Guides\RestructuredText\Directives;
 use phpDocumentor\Guides\Nodes\Node;
 use phpDocumentor\Guides\RestructuredText\Parser\Directive;
 use phpDocumentor\Guides\RestructuredText\Parser\DocumentParserContext;
-use phpDocumentor\Guides\RestructuredText\TextRoles\DefaultTextRoleFactory;
+use phpDocumentor\Guides\RestructuredText\TextRoles\TextRoleFactory;
 
 /**
  * sets the default interpreted text role, the role that is used for interpreted text without an explicit role.
@@ -26,7 +26,7 @@ use phpDocumentor\Guides\RestructuredText\TextRoles\DefaultTextRoleFactory;
 class DefaultRoleDirective extends BaseDirective
 {
     public function __construct(
-        private readonly DefaultTextRoleFactory $textRoleFactory,
+        private readonly TextRoleFactory $textRoleFactory,
     ) {
     }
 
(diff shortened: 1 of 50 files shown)


packages/guides-restructured-text/src/RestructuredText/Directives/DefaultRoleDirective.php
```diff
@@ -0,0 +1,47 @@
...
+        private readonly TextRoleFactory $textRoleFactory,
+    ) {
+    }
+
+    public function getName(): string
+    {
+        return 'default-role';
+    }
+
+    public function process(
```
jaapio, line 38: `processAction` was intended for this type of directives. However I noticed that the base directive never calls this method, and it seems never used. Maybe we should consider to enable this usage, by introducting a new base class, that will provide an method like `processAction`

```
abstract class ActionDirective 
{
   process(): {
      $this->processAction();
      return null;
   }
   
    public function processAction(
        DocumentParserContext $documentParserContext,
        string $variable,
        string $data,
        array $options,
    ): void {
    }
}

File: packages/guides-restructured-text/src/RestructuredText/TextRoles/DefaultTextRoleFactory.php [hunk 2]
```diff
@@ -17,19 +17,37 @@ class DefaultTextRoleFactory implements TextRoleFactory
...
         private array $domains = [],
     ) {
         $this->textRoles = [...$textRoles];
     }
 
-    public function registerTextRole(TextRole $textRoles): void
+    public function registerTextRole(TextRole $textRole): void
     {
-        $this->textRoles[] = $textRoles;
+        $this->textRoles[] = $textRole;
```
jaapio, line 29: Should we not create an name based map to store the text roles? As this is there unique identifier?

jaapio (changes requested):
I could not find this in your changes to that's why I ask it, default text roles set by a document are document bases, aren't they? Otherwise the order of parsing matters and I don't thing we have that order currently. It would also disallow caching of the full documents between runs as the parser would be in a different state. Therefor I think it would make sense to add these textroles somehow in the DocumentParserContext? 

Same applies for document defined or replaced roles, those should only apply to the current document from the point where the role was defined.

packages/guides-restructured-text/src/RestructuredText/TextRoles/DefaultTextRoleFactory.php
linawolf: One testrole can have multiple aliases

packages/guides-restructured-text/src/RestructuredText/Directives/DefaultRoleDirective.php
linawolf: Good idea

linawolf:
We usually define textroles in the includes.rst.txt which is included at the beginning of each document. So I supose it is probably defined on document level. will have a look at it

packages/guides-restructured-text/src/RestructuredText/Directives/DefaultRoleDirective.php
linawolf: I introduced the abstract class and extended it

linawolf:
@jaapio Each DocumentParserContext now manages a clone of the TextRoleFactory such that changes made to it by one directive do not apply to another document. Also added a test for this.

jaapio (approved):
Thanks this looks good!

PR #534, opened 2023-08-12
[FEATURE] introduce highlight directive

linawolf:
to set default languages for code-blocks https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-highlight

depends on #538


jaapio:
@linawolf  can you please rebase this PR? So we can continue with reviews and merging :-D

jaapio (approved):

PR #537, opened 2023-08-13
[FEATURE] Mark active and current menu entries

linawolf:


packages/guides/src/Compiler/NodeTransformers/MenuNodeAddSubDocumentsTransformer.php
```diff
@@ -51,16 +52,20 @@ private function addSubEntries(
...
         foreach ($documentEntry->getChildren() as $subDocumentEntryNode) {
             $subMenuEntry = new MenuEntryNode(
                 $subDocumentEntryNode->getFile(),
                 $subDocumentEntryNode->getTitle(),
                 [],
                 false,
                 $currentLevel,
+                '',
+                self::isInRootline($subDocumentEntryNode, $compilerContext->getDocumentNode()->getDocumentEntry()),
+                self::isCurrent($subDocumentEntryNode, $currentPath),
```
jaapio, line 65:
```suggestion
                self::isCurrent($subDocumentEntryNode, $compilerContext->getDocumentNode()->getFilePath()),
```

packages/guides/src/Compiler/NodeTransformers/MenuNodeAddSubDocumentsTransformer.php
```diff
@@ -69,6 +74,18 @@ public function supports(Node $node): bool
         return $node instanceof TocNode || $node instanceof NavMenuNode;
     }
 
+    private function isInRootline(DocumentEntryNode $menuEntry, DocumentEntryNode $currentDoc): bool
+    {
+        return $menuEntry->getFile() === $currentDoc->getFile()
```
jaapio, line 79: I would like to split this line into multiple return statements. This will make the code more readable. 

if (...)  {
   return true;
}

if ($current->getParent() === null) {
   return false;
}

return false;

jaapio (approved):
Just a few small remarks. But this looks good.

linawolf:

jaapio (approved):

PR #538, opened 2023-08-13
!!![TASK] Move LineIterators to Block Context

linawolf:
This is quite a breaking change. However we had a real flaw in design here. We stored blocks of lines as line iterator in the DocumentParserContext that do not represent the document. To prevent adverse effects we cloned the DocumentParserContext. But this means that all data stored in the document context can get lost because it is set in a clone and not the original context.

This is needed to make #534 work


jaapio (approved):
Thanks, this looks very good!

PR #546, opened 2023-08-20
[BUGFIX] Make underscores work as first sign in a code role

linawolf:
This combination was interpreted as the end of a named reference, we were tokenizing to lax here.

PHPstan does not get that $lexer->moveNext() changes the content of $lexer->token, therefore instead of having to apply many quirks I generated a phpstan-baseline. I think the issue should be fixed in the lexer library by making $token private and a getToken method instead.


jaapio (commented):
Sorry, I should read the description

jaapio (approved):

PR #547, opened 2023-08-21
[FEATURE] Introduce Breadcrumb

linawolf:
The breadcrumb can either be inserted as directive of that name or directly into the template by using the twig function {{ renderBreadcrumb() }}. See Bootstrap theme as example for the latter


jaapio (commented):
There seems to be quiet some logic in the renderer. It would have been nice if we could move some of this logic to the compiler. This would be more in line with the way toctree and menu are working. 

If you see it from an abstract level a breadcrumb path could be some flat menu structure where every node has a single child. Given the fact toctrees have an impact on the structure of documents, I think it is wise to use this same structure in breadcrumbs.

packages/guides/src/Nodes/BreadCrumbNode.php
```diff
@@ -0,0 +1,14 @@
...
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Nodes;
+
+/** @extends AbstractNode<string> */
+class BreadCrumbNode extends AbstractNode
+{
+    public function __construct(string $value = '')
+    {
+        $this->value = $value;
```
jaapio, line 12: Please call the parent constructor, this prevents issues when we change the parent class.

linawolf: Calling the parent constructor on a class that has no constuctor is not possible...

packages/guides/src/NodeRenderers/Html/BreadCrumbNodeRenderer.php
```diff
@@ -0,0 +1,128 @@
...
+            $documentEntry->getTitle(),
+            [],
+            false,
+            $level,
+            '',
+            true,
+            $isCurrent,
+        );
+        $currentBreadcrumb[] = $entry;
+        if ($documentEntry->getParent() === null) {
```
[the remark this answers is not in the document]
linawolf: I kind of prefere to have the expected return as last return in the method while the special cases are handled first... Your change would revert that order.

jaapio (approved):

PR #548, opened 2023-08-21
[Feature]: Add csv table directive

jaapio:
Add csv table directive, this allows users to import a csv as table.

### DIFF
--- a/tests/Integration/tests/table-csv-table-with-markup/input/test.csv
+++ b/tests/Integration/tests/table-csv-table-with-markup/input/test.csv
@@ -1,0 +1,4 @@
+col 1, col 2
+**a**, "quoted column is unquoted"
+c, d
+, first column is empty
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "minimum-stability": "stable",
     "config": {
         "platform": {
-            "php": "8.1.0"
+            "php": "8.1.2"
         },
         "allow-plugins": {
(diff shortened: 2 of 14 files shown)


linawolf (approved):
I would suggest to maybe add a few more complex test cases

linawolf:
Also note that there is such a feature in rst: https://docutils.sourceforge.io/docs/ref/rst/directives.html#csv-table-1, we also use it in TYPO3 https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Content/Tables.html#csv-table

tests/Integration/tests/table-csv-table-with-markup/input/test.csv
```diff
@@ -0,0 +1,4 @@
+col 1, col 2
+**a**, "quoted column is unquoted"
```
linawolf, line 2: I am not sure this makes sense in most use cases... Often csv files would come from others sources so I would not really expect rst markup to be applied? What you think?

jaapio: According to this page: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html#csv-files

it is supported by sphinx?

linawolf: Ok, didnt know that

PR #553, opened 2023-08-23
!![TASK] Replace the subparser with a starting rule in directives

linawolf:
This change is breaking for anyone who implemented a SubDirective with a constructor as now the parent constructor has to be called. It also fixes the bug of disappearing field lists in direcitves.

getSubParser is only used in 2 other places that can be dealed with in a follow up such that we can remove the deprecated method all together,

I also removed the subparser in the Blockquotes. So it is only used in the include directive now.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveContentRule.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveContentRule.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\RestructuredText\Parser\Productions;
+
+use phpDocumentor\Guides\Nodes\CompoundNode;
+use phpDocumentor\Guides\Nodes\Node;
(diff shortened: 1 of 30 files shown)


packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DirectiveContentRule.php
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public function applies(BlockContext $blockContext): bool
+    {
+        return true;
+    }
+
+    public function apply(BlockContext $blockContext, CompoundNode|null $on = null): Node|null
+    {
+        $node = new DirectiveContentNode([]);
```
jaapio, line 26: We could just use the collectionnode here as the directivecontentnode is not an existing node. It's just for parsing collection. Most directives are directly copying the children from this class.

linawolf: Good idea!

jaapio (approved):

PR #554, opened 2023-08-23
Endorse divergence from specification for links

greg0ire:
According to
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup, inline markup must be separated from surrounding text by non-word characters.

Currently, the library does not seem to be checking for the spaces, and it seems that Github's parser has the same behavior: a link between brackets is recognized.

That behavior is nice, and since Github adopted it, it might not be that risky. Let us add a test so that it can be relied on.

### DIFF
--- a/tests/Functional/tests/links/links.html
+++ b/tests/Functional/tests/links/links.html
@@ -9,3 +9,4 @@
 pick it up from <a href="ftp://foo.example.com/rfc/">ftp://foo.example.com/rfc/</a>. Note the warning in
 <a href="http://www.ics.uci.edu/pub/ietf/uri/historical.html#WARNING">http://www.ics.uci.edu/pub/ietf/uri/historical.html#WARNING</a>.</p>
 <p>You can read more on the features of Embeddables objects <a href="http://docs.doctrine-project.org/en/latest/tutorials/embeddables.html">in the documentation</a>.</p>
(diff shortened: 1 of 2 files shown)


linawolf (approved):

wouterj:
Fyi, this is not a divergence from [the specification](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules) as far as I know:

> Inline markup start-strings must start a text block or be immediately preceded by
>
>  * whitespace,
>  * one of the ASCII characters - : / ' " < ( [ {
>  * or a similar non-ASCII punctuation character. [19]
>
> Inline markup end-strings must end a text block or be immediately followed by
>
>  * whitespace,
>  * one of the ASCII characters - . , : ; ! ? \ / ' " ) ] } >
>  * or a similar non-ASCII punctuation character. [20]
>
>  [19] Unicode categories Ps (Open), Pi (Initial quote), Pf (Final quote), Pd (Dash), or Po (Other).
> [20] Unicode categories Pe (Close), Pi (Initial quote), Pf (Final quote), Pd (Dash), or Po (Other).

greg0ire:
Oh right indeed! Well that's a bug in `doctrine/rst-parser` then 👍

PR #556, opened 2023-08-25
[TASK] Replace the subparser in the include directive

linawolf:


linawolf:
@jaapio I cannt seem to make this work in the include directive. Not sure why. We might need the subparser concept here for some reason?

jaapio:
The issue you might be facing is that the injected rule is not a shared instance but an empty new instance? 

----
Aside from that, I'm not entirely sure how this should work, as I understood the `include` directive it does some copy past magic in the background to literally make the included document part of it's parent? For us this means that the `DocumentParserContext` will be the same for the included document as it was for it's parent.

PR #562, opened 2023-08-26
[FEATURE] Introduce quote like body directives

linawolf:
https://docutils.sourceforge.io/docs/ref/rst/directives.html#epigraph https://docutils.sourceforge.io/docs/ref/rst/directives.html#highlights https://docutils.sourceforge.io/docs/ref/rst/directives.html#pull-quote


packages/guides-restructured-text/src/RestructuredText/Directives/EpigraphDirective.php
```diff
@@ -0,0 +1,36 @@
...
+
+    /** {@inheritDoc}
+     *
+     * @param Directive $directive
+     */
+    protected function processSub(
+        CollectionNode $collectionNode,
+        Directive $directive,
+    ): Node|null {
+        return new QuoteNode($collectionNode->getChildren(), ['epigraph']);
```
jaapio, line 34: Should we not add an extra class to the last paragraph node to mark it as attribution?

jaapio (approved):

packages/guides-restructured-text/src/RestructuredText/Directives/EpigraphDirective.php
linawolf: The last paragraph is only an attribution if it starts with two -- so we have to add an extra test for that then

jaapio: Maybe an extra rule for that would make sense? that is only applied to this specific directive. As you might want to remove the `--` starter from the final node as well? 

Let's do that in a followup PR, because this directive needs more attention :-)

PR #565, opened 2023-08-26
[BUGFIX] Replace special signs in directive template names

linawolf:
And handle directives in namespaces as having templates in subdirectories


jaapio (commented):
A unittest for this would be nice. To ensure we do not break this in the future.

linawolf:
@jaapio I added tests, please have another look

jaapio (approved):

PR #573, opened 2023-09-03
[BUGFIX] Make Email detection in lexer more strict

linawolf:
So it does not catch email like expressions in backticks


packages/guides-restructured-text/src/RestructuredText/Parser/InlineLexer.php
```diff
@@ -56,7 +56,7 @@ protected function getCatchablePatterns(): array
         return [
             '\\\\``', // must be a separate case, as the next pattern would split in "\`" + "`", causing it to become a intepreted text
             '\\\\[\s\S]', // Escaping hell... needs escaped slash in regex, but also in php.
-            '\\S+@\\S+\\.\\S+',
+            '[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}',
```
jaapio, line 59: This is good enough for now... but it feels like this should be done in a different way... as emailaddreses are not easy to catch in a regex.

PR #575, opened 2023-09-03
[FEATURE] Make csv-table directive work with content

linawolf:


jaapio (commented):
It feels like we would need to introduce some extra layer here, some kind of strategy that can be executed per variant.  

I don't know if I'm making things over complex by this but I would like to have your opinion on this. 

```php
interface CsvTableFactory {
   public function supports(Directive $directive, BlockContext $blockContext): bool;
   public function create(Directive $directive, BlockContext $blockContext): Node; 
} 
```

```php
class CsvTableStringFactory implements CsvTableFactory 
{
   public function supports(Directive $directive, BlockContext $blockContext): bool {
      return count($blockContext->getDocumentIterator()->toArray()) > 0
   }
   
   public function create(Directive $directive, BlockContext $blockContext) {
      /*...*/
      
      return new TableNode($rows, array_filter([$header]));
   }
}
```
This makes the code cleaner to me... and would allow us to be more strict on the combinations of options that are used... for example... when `:file:` is set, we ignore the content, but also the `:header:`

linawolf:
I think my PR is working as is and it is a seldom used directive for us. Right now calling csv directives causes a file not found error. If you want to apply your ideas in a follow up go ahead. There are however much more frequently used directives where we have no check for silently

PR #578, opened 2023-09-03
[FEATURE] Introduce twig function to output menu

linawolf:
This way in the bootstrap template we can always output a menu even if no block is defined


linawolf:
@jaapio Could you have a look please? I need this to move forward with my docs rendering

packages/guides/src/Nodes/DocumentNode.php
```diff
@@ -201,6 +205,17 @@ public function hasDocumentPart(string $identifier): bool
         return isset($this->documentPartNodes[$identifier]);
     }
 
+    public function addTocNode(TocNode $tocNode): void
```
jaapio, line 208: Why do we need this in a separate array? Is the function getNodes not enough to collect them from the tree?

packages/guides/src/Nodes/ProjectNode.php
```diff
@@ -118,6 +118,17 @@ public function getAllDocumentEntries(): array
         return $this->documentEntries;
     }
 
+    public function getRootDocumentEntry(): DocumentEntryNode
+    {
+        foreach ($this->documentEntries as $documentEntry) {
```
jaapio, line 123: If there is just one root node, it might be better to keep this as a property of the project node. 
And remove the isRoot from the document. 

Otherwise this method should return all root documents.

packages/guides/src/RenderContext.php
```diff
@@ -30,6 +31,8 @@ class RenderContext
     private string $destinationPath;
 
     private DocumentNode $document;
+    /** @var DocumentNode[] */
+    private array $allDocuments;
```
jaapio, line 35: Would it be an idea to collect all documents in the projectnode and pass that in the renderingcontext?

jaapio (changes requested):

packages/guides/src/RenderContext.php
linawolf: This does not work because someone :tm: wants the nodes to be immutable

packages/guides/src/Nodes/ProjectNode.php
linawolf: There is just one. The implementation details do not worry so much I think. We do not have so many documents that it would matter if we search for it once in a while?

packages/guides/src/Nodes/DocumentNode.php
linawolf: The thing is that the tocnodes might be in a section which is in a section which is in a section, so you have to traverse the whole tree to search for them, for which we would need to introduce several new methods

jaapio (approved):

PR #586, opened 2023-09-09
[FEATURE] Configure inventories via guides.xml

linawolf:


jaapio (approved):

packages/guides/src/DependencyInjection/GuidesExtension.php
```diff
@@ -39,6 +39,18 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->scalarNode('version')->end()
                     ->end()
                 ->end()
+                ->arrayNode('inventories')
```
wouterj, line 42: You could also make this mode a prototyped array node. And add `->fixXmlConfig('inventory', 'inventories')` (see the 4th example in https://symfony.com/doc/current/components/config/definition.html#array-nodes ).

This creates a config like

```xml
<guides>
  <inventory>...</inventory>
  <inventory>...</inventory>
</guides>
```

And produces an array `['inventories' => ['...', '...']]`

This is the common approach used by Symfony bundles when dealing with prototype nodes.

PR #587, opened 2023-09-10
[FEATURE] Set console option defaults via guides.xml

linawolf:


guides.xml
```diff
@@ -3,6 +3,16 @@
...
+    <html_theme>bootstrap</html_theme>
+    <input>docs</input>
+    <output>output</output>
+    <input_format>rst</input_format>
+    <output_formats>
+        <output_format>html</output_format>
+    </output_formats>
+    <log_path>php://stder</log_path>
+    <fail_on_log>false</fail_on_log>
+    <show_progress>true</show_progress>
```
wouterj, line 15: In xml, it's allowed to replace underscores with dashes, which is much more common in XML attribute and element names (e.g. `fail-on-log`).

Also, it's common to use attributes instead of elements when dealing with scalar nodes (instead of array ones): `<guides fail-on-log="false">`

linawolf: However the existing tags base_template_paths and html_theme also use underscores?

linawolf: I tried renaming them but as soon as I rename them they do not work anymore and I get:

PHP Fatal error:  Uncaught Symfony\Component\Config\Definition\Exception\InvalidConfigurationException: Unrecognized option "input_format" under "guides". Did you mean "input-for
mat"? in /var/www/html/vendor/symfony/config/Definition/ArrayNode.php:327
Stack trace:
#0 /var/www/html/vendor/symfony/config/Definition/BaseNode.php(385): Symfony\Component\Config\Definition\ArrayNode->normalizeValue(Array)
#1 /var/www/html/vendor/symfony/config/Definition/Processor.php(32): Symfony\Component\Config\Definition\BaseNode->normalize(Array)

wouterj: Sorry, it's the other way around. You use `->scalarNode('input_format')` in the configuration builder, but `input-format` in the XML code.
The xml loader will automatically replace the dashes with underscores before processing it.

linawolf:
@wouterj I think I understood it, please have another look?

wouterj (approved):
Looks good!

PR #592, opened 2023-09-16
[BUGFIX] Make setting theme in guides.xml work

linawolf:
Until now the theme was set before all themes where loaded causing an error because the theme was not found.


packages/guides-cli/resources/schema/guides.xsd
```diff
@@ -14,7 +14,7 @@
             <xsd:element name="extension" type="extension" minOccurs="0" maxOccurs="unbounded"/>
         </xsd:choice>
 
-        <xsd:attribute name="html-theme" type="xsd:string"/>
+        <xsd:attribute name="html_theme" type="xsd:string"/>
```
jaapio, line 17: Would `theme` not be a better name? As the theme doesn't have to be html?

linawolf: The setting already existed, I only made it work. But its fine to me to rename it, then it has the same name like the parameter of the cli, which is probably more logig

jaapio (approved):

packages/guides-cli/resources/schema/guides.xsd
wouterj: Yes, we already use themes/theme to configure which themes are available. As this is mostly an internal thing, we can probably rename it to a more complex name.
The html_theme name is inspired by Sphinx. This allows to use different themes for different output formats, if we ever add support for more (e.g. PDF export with a different view from the HTML).


Btw, as I also said in the other PR, the dash version (`html-theme`) should work and be preferred. I'm afk at the moment, but I'm pretty sure it works.

jaapio: Sorry I missed this comment

PR #598, opened 2023-09-23
[TASK] Make basic latex output work

### LINKED ISSUE #444: Fix Latex output
Skipped test: https://github.com/phpDocumentor/guides/tree/main/tests/Integration/tests-latex/index

linawolf:
For now this is only a proof of concept. to really produce useful latex output more work will be needed.

resolves #444


jaapio:
However this might work I do not really like the way this is implemented right now. My idea was to have node renderers registered per output format. You can see this in: https://github.com/phpDocumentor/guides/blob/main/packages/guides/src/DependencyInjection/Compiler/NodeRendererPass.php#L144

For latex a new set of renders should be registered, some of them will use the same implementation as the html ones, or just render a twig template, while others won't. 
To make this work we have to register a second `\phpDocumentor\Guides\Handlers\RenderDocumentHandler` that gets the Latex document node renderer passed as a renderer, From there the whole flow will start rolling. 

I understand that this is a lot more work than your current change, but it will provide us more flexibility. I would understand if you would simply let this go for now.

linawolf:
Let us postphone this then

PR #599, opened 2023-09-24
[FEATURE] Add option input-file to render a specific file

linawolf:
This can be used to parse READMEs or other standalone reST files. It is also a preparation for rendering specific mark down files in future like a README.md

### DIFF
--- a/tests/Integration/tests/input-file/input/guides.xml
+++ b/tests/Integration/tests/input-file/input/guides.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides packages/guides-cli/resources/schema/guides.xsd"
+        input-file="README.rst"
+>
+</guides>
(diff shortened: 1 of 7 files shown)


tests/Integration/tests/input-file/input/guides.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides packages/guides-cli/resources/schema/guides.xsd"
+        input-file="README.rst"
```
jaapio, line 5: I do understand this use case on the commandline, but is it required to have this in the configuration file? That doesn't really make sense to me.

linawolf: Where is the Harm of having it in the configuration file as we have basically all settings that you can do in the command line also in the configuration file, why should we divide from that here?

linawolf: @jaapio how shall we resolve this?

jaapio (approved):

PR #600, opened 2023-09-24
[TASK] Make Markdown availible in integration tests

linawolf:
Install the markdown extension in the composer.json and load it for the integration tests. Introduce a test (still skipped as there are issues to solve)


linawolf:
@jaapio I think you can use this as base for making MD work

jaapio (approved):

jaapio:
Thanks!

PR #603, opened 2023-09-26
Feat: Add plantuml server rendering

jaapio:
This patch makes is possible to load configuration for extensions that are not loaded by default. This gives us the option to add configuration to the graph extension.

By now this extension supports the options to use an remote plantuml server to render uml diagrams.


packages/guides-graphs/resources/config/guides-graphs.php
```diff
@@ -3,15 +3,37 @@
...
+
+        ->set(PlantumlRenderer::class)
+        ->arg('$plantUmlBinaryPath', '%guides.graphs.plantuml_binary%')
+
+        ->set(PlantumlServerRenderer::class)
+        ->arg(
+            '$plantumlServerUrl',
+            '%guides.graphs.plantuml_server%',
+        )
+        ->alias(DiagramRenderer::class, PlantumlServerRenderer::class)
```
jaapio, line 35: For now this is not configurable, the configuration options are there, but I need to add a service locator that allows you to register any class.

packages/guides-cli/src/DependencyInjection/ContainerFactory.php
```diff
@@ -54,7 +54,7 @@ public function loadExtensionConfig(string $extension, array $config): void
 
         $extensionAlias = $this->registeredExtensions[$extensionFqcn] ?? false;
         if (!$extensionAlias) {
-            $this->registerExtension(new $extensionFqcn());
+            $this->registerExtension(new $extensionFqcn(), $config);
```
jaapio, line 57: @wouterj I think you missed this when implementing the extension config? Otherwise only default extensions can be configured, not the ones configured in the guides.xml?

wouterj: You're right :+1: 

I think I originally intended this without the early return on the next line (so we "register" the extension without config, and then add extra config). But this works as well.

linawolf (approved):
I think it is a good start!

linawolf:
@jaapio should we merge this? Further open issues can be adressed afterwards

PR #608, opened 2023-10-01
Allow setting a default language for literal blocks

wouterj:
Besides setting the local language using the `.. highlight::` directive, Sphinx also allows specifying a global default language for literal blocks. We use this in Symfony to mark all literal blocks as "php".

After this PR, you can configure the default language in `guides.xml`:

```xml
    <extension class="phpDocumentor\Guides\RestructuredText"
        default-code-language="php"
    />
```

### DIFF
--- a/packages/guides-cli/resources/schema/guides.xsd
+++ b/packages/guides-cli/resources/schema/guides.xsd
@@ -23,6 +23,7 @@
         <xsd:attribute name="fail-on-log" type="xsd:string"/>
         <xsd:attribute name="show-progress" type="xsd:string"/>
         <xsd:attribute name="theme" type="xsd:string"/>
+        <xsd:attribute name="default-code-language" type="xsd:string"/>
     </xsd:complexType>
 
     <xsd:complexType name="extension">
--- a/packages/guides-restructured-text/resources/config/guides-restructured-text.php
+++ b/packages/guides-restructured-text/resources/config/guides-restructured-text.php
@@ -53,6 +53,7 @@ use phpDocumentor\Guides\RestructuredText\Directives\VersionChangedDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\WarningDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\WrapDirective;
 use phpDocumentor\Guides\RestructuredText\MarkupLanguageParser;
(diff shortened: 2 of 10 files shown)


packages/guides-restructured-text/src/RestructuredText/DependencyInjection/ReStructuredTextExtension.php
```diff
@@ -21,13 +25,34 @@
...
 
+    public function getConfigTreeBuilder(): TreeBuilder
+    {
+        $treeBuilder = new TreeBuilder('restructured_text');
+        $rootNode = $treeBuilder->getRootNode();
+        assert($rootNode instanceof ArrayNodeDefinition);
+
+        $rootNode
+            ->children()
+                ->scalarNode('default_code_language')->defaultValue('')->end()
```
greg0ire, line 40: Once this is merged, I suppose #475 should use that parameter?

wouterj: No, it's works automatically: `CodeNode`'s have a `language` property. If it's not explicitly set (e.g. by the `.. code-block:: html` directive), it'll default to this config property.

The highlighter can just use `codeNode.language`.

linawolf:
We do save the language in each LiteralInlineNode, which is already part of the guides. So why don't we make this part of the guides and Save it in the ProjectSettings?  It is a useful feature for other input formats as well.

Also I think we should keep all Settings in the Settings Manager so that they are easy to find and not scattered throughout the project. We could make a way that extensions can save their setting objects in the settings Manager?

wouterj:
Good suggestions, I've moved the setting to `ProjectSettings`/guides.

linawolf:
I would suggest to add an integration test that demonstrates this setting. Integration tests do respect the guides.xml

linawolf:
I am going to rebase this and see what I can do about the integration tests

linawolf:
@wouterj could you have a look why the integration tests are not working?

linawolf:
Ah ok, I think I figured it out. it needs to be set as an argument of the guides tag now, however we need to change the description of your commit and the guides.xsd

linawolf (approved):

PR #614, opened 2023-10-03
[FEATURE] Make relative, absolute and links relative to a base URL possible

linawolf:
depends on #607


jaapio (changes requested):
Thanks for this massive improvement!

In general this looks very good, however I would like to see an improvement that makes the UrlGenerator a bit more light weight. And that might reduce the number of changes you need to do.

Right now the UrlGenerator is basically handling 2 different strategies:
1. Generate urls relative
2. Generate urls absolute

If we would split this into 2 different classes, we could reduce the complexity of the UrlGenerator and also make it easier to understand.
Beside that you could replace the `new UrlGenerator` in the RenderContext creation with an 
injected UrlGeneratorInterface. This would make it easier to replace the UrlGenerator with a custom implementation.

This way the RenderContext does not require the settingsManager anymore. And we could
even inject the UrlGenerator everywhere we need it. To accomplish this
you could set the correct UrlGenerator in the GuidesExtension::load method, like you 
do with the ThemeManager.

Another benefit of this approach is that we could inverse the dependency on the UrlGenerator
and RenderContext. This way the UrlGenerator can take what ever it needs from the RenderContext.
For example:

```php
   $urlGenerator->generateUrl($renderContext, $cannonicalUrl);
```

As an extra, you could allow people to configure the UrlGeneator in the Guides.xml. 

```xml
<guides>
    <urlGenerator class="My\Custom\UrlGenerator" />
</guides>
````
 
Or for buildin strategies:

```xml
<guides>
    <urlGenerator strategy="relative" />
</guides>
```

packages/guides/tests/unit/Renderer/UrlGenerator/RelativeUrlGeneratorTest.php
```diff
@@ -0,0 +1,115 @@
...
+                'filename' => 'file',
+                'outputFormat' => 'html',
+                'anchor' => 'anchor',
+            ],
+            'Empty File with anchor' => [
+                'expected' => '#anchor',
+                'filename' => '',
+                'outputFormat' => 'html',
+                'anchor' => 'anchor',
+                'skip' => 'Empty filenames are not supported',
```
jaapio, line 104: can we support this in the future? or is this an error situation?

jaapio (approved):
Just a single question. This looks good to go for me.

Thanks!

packages/guides/tests/unit/Renderer/UrlGenerator/RelativeUrlGeneratorTest.php
linawolf: Yes sure. Just vheck if the document name is empty?

PR #626, opened 2023-10-17
Fix: allow absolute image urls

### LINKED ISSUE #621: RST `image` directive tries to load images from local, even if explicitly passed as absolute URLs
I am using https://pandoc.org (which, heck, I wish I knew it earlier) to generate `rst` from my markdown (and other) file formats.
This works like a charm.

However, there is a peculiarity in the PHPDocumentor `Guides` feature.
Whenever I generate a Guide using the `rst` file it generates `img` tags that look for the image source file on the _local_ path. That of course causes the images to not load.

**Example**

- In the `rst` you have:
```
|GitHub contributors|
// other content
.. |GitHub contributors| image:: https://img.shields.io/github/contributors/ORG/REPO
```
- PHPDocumentor Guide HTML output:
```
<img src="guide/tukutoi-template-builder/https://img.shields.io/github/contributors/ORG/REPO">
```

It is self-explanatory why I cannot use _local_ files here: the whole point of this badge is that it is dynamically generated by GitHub.

Can this be fixed somehow? I already tried to wrap my absolute URL in `<>` in the `rst` file, which did not help.
Perhaps some config attribute, or terminal command I am not aware of?

jaapio:
To be able link to external urls we need to check at multiple places how to handle the url.

fixes #621

### DIFF
--- a/packages/guides/resources/template/html/body/image.html.twig
+++ b/packages/guides/resources/template/html/body/image.html.twig
@@ -1,5 +1,5 @@
 <img
-    src="{{ asset(node.value) }}"
+    src="{%- if node.value is external_target -%} {{ node.value }} {%- else -%} {{ asset(node.value) }} {%- endif -%}"
     {% if node.hasOption('width') %}width="{{ node.option('width') }}"{% endif%}
     {% if node.hasOption('height') %}height="{{ node.option('height') }}"{% endif%}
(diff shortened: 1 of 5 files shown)


linawolf (approved):
Looks good to me, didnt even know, that was possible

PR #634, opened 2023-10-25
Highlighter improvements

wouterj:
Couple fixes while testing out the new guides-code extension in the Symfony docs.

### DIFF
--- a/packages/guides-code/src/Code/Highlighter/Highlighter.php
+++ b/packages/guides-code/src/Code/Highlighter/Highlighter.php
@@ -47,13 +47,6 @@ final class Highlighter
             $highlightValue = $highlight->value;
             $highlightLanguage = $highlight->language;
 
-            if ($language === 'terminal') {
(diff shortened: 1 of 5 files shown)


packages/guides-code/src/Code/Highlighter/Highlighter.php
```diff
@@ -47,13 +47,6 @@ public function __invoke(string $language, string $code): HighlightResult
             $highlightValue = $highlight->value;
             $highlightLanguage = $highlight->language;
 
-            if ($language === 'terminal') {
```
wouterj, line 50: I see we have tests for this behavior. To me, this is very specific to how we want to show terminal blocks in the Symfony docs (so I'm happy to keep it in our project). Or do you think this is generic enough to keep in this project?

jaapio: it makes sense to me to remove this, and move this to your own project, this seems to be very specific indeed.

jaapio:
It would be nice to have some small notes somewhere how to use this extension. Maybe based on your implementation? But that's not blocking me from merging this. 

Thanks!

PR #627, opened 2023-10-21
[FEATURE] Introduce an Anchor reduction strategy

linawolf:
The AnchorReducer can be implemented by applications if different handling is needed.

Anchors are stored in project context now (as they should work project wide) however they do not work in functional tests anymore.

We might want to reconsider workings of the functional tests...

resolves #617

PR #636, opened 2023-10-26
[FEATURE] Introduce option directive

linawolf:
Options can be linked with the special text-role :option:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option

depends on #627


packages/guides/src/Nodes/SectionNode.php
```diff
@@ -33,4 +35,19 @@ public function getTitles(): array
...
+
+    public function getId(): string
+    {
+        return $this->getTitle()->getId();
+    }
+
+    public function getLinkText(): string
+    {
+        return $this->getTitle()->toString();
+    }
```
jaapio, line 52: Is this not rendering information? A SectionNode can be rendered as a "LinkTarget" or as a "Section", in the first case we use a specific rendering strategy to transform the SectionNode to a link. 

This will make the SectionNode unaware of the contexts using the node. The node itself doesn't have to be aware of this. Which makes it easier to maintain this node or introduce new nodes with the same way of rendering.

packages/guides/src/ReferenceResolvers/AnchorReferenceResolver.php
```diff
@@ -24,7 +25,12 @@ public function __construct(
     public function resolve(LinkInlineNode $node, RenderContext $renderContext): bool
     {
         $reducedAnchor = $this->anchorReducer->reduceAnchor($node->getTargetReference());
-        $target = $renderContext->getProjectNode()->getInternalTarget($reducedAnchor);
+        if ($node instanceof ReferenceNode) {
+            $target = $renderContext->getProjectNode()->getInternalTarget($reducedAnchor, $node->getLinkType());
+        } else {
+            $target = $renderContext->getProjectNode()->getInternalTarget($reducedAnchor);
+        }
+
```
jaapio, line 33: It feels a bit odd to have a else in here, as we already have a way to find the correct Reference resolvers. Should this not be a different strategy? maybe with some shared base class, but right now this looks like a strategy inside a strategy. 

If I look at the example of the in phpDocumentor introduced link targets, I use the domain to check if the domain is `php` if it's not a known `php` link I ignore the link and pass it on to the next resolver. I think this should also be the case for other `LinkTypes`.

packages/guides/src/Nodes/ProjectNode.php
```diff
@@ -91,17 +91,21 @@ public function getCitationTarget(string $name): CitationTarget|null
         return $this->citationTargets[$name] ?? null;
     }
 
-    public function addLinkTarget(string $anchorName, InternalTarget $target): void
+    public function addLinkTarget(string $anchorName, InternalTarget $target, string $linkType = SectionNode::STD_LABEL): void
```
jaapio, line 94: Why is `$linkType` not part of the `$target`? It's an property of the `$target` telling me something about what kind of target it is. This would make the search a bit more complex, however we can solve that by creating a collection class that is responsible for the separation of link types. 

This will make the project node less specific for this use case.

jaapio (changes requested):
It feels like the term intersphinxDomain is invalid or should not be part of our core library. I'm still thinking if and how intersphinx should be part of our core.  However domains are a more common thing in rst, at least it sounds less sphinx specific. Which makes sense to me to rename the current `getIntersphinxDomain` to `getDomain` 

I think the `:option:` domain should have it's own `ReferenceResolver`, it's another type of reference, so let's model it like that. This ReferenceResolver should be part of the `Rst` package as it is not a shared concept between RST and Markdown for example. Probably this is also the reason why you have put the node inside the RST extension.

packages/guides/src/Nodes/SectionNode.php
linawolf: What I was doing here was introducing a "LinkTargetNode"  interface that can be implemented by any node that can be linked to. Now anything that can be linked to supplys a default link text that can be used when no explicit linktext is supplied by the link. This link text is not used for rendering the node itself but any link to it. By making also the Section Node implement the interface the section node itself can be treated like any other linkable object.

packages/guides/src/ReferenceResolvers/AnchorReferenceResolver.php
linawolf: I added a todo, we can solve this later

jaapio (approved):

PR #640, opened 2023-10-28
[FEATURE] Enable basic Markdown files

linawolf:
Most basic block and inline formats are covered now.

Markdown - different from rst - allows nested inline elements (a strong text that is also emphasized, link texts with inline code, ...) Our nodes can currently not store such information therefore we cannot support them currently.

Images are considered inline in Markdown and can appear within texts while images in rst are block level elements. I therefore had to introduce a new Node Type for inline images.

League\CommonMark Does not seem to support line-breaks (https://www.markdownguide.org/basic-syntax/#line-breaks) so these are also missing.

I did not implement https://www.markdownguide.org/basic-syntax/#html due to the security concerns.

### DIFF
--- a/packages/guides-markdown/src/Markdown/Parsers/BlockQuoteParser.php
+++ b/packages/guides-markdown/src/Markdown/Parsers/BlockQuoteParser.php
@@ -49,7 +49,7 @@ final class BlockQuoteParser extends AbstractBlockParser
                 return new QuoteNode($content);
             }
 
-            $this->logger->warning(sprintf('BlockQuote CONTEXT: I am leaving a %s node', $commonMarkNode::class));
+            $this->logger->warning(sprintf('"%s" node is not yet supported in context %s. ', $commonMarkNode::class, 'BlockQuote'));
         }
 
         throw new RuntimeException('Unexpected end of NodeWalker');
--- a/packages/guides-markdown/src/Markdown/Parsers/BlockQuoteParser.php
+++ b/packages/guides-markdown/src/Markdown/Parsers/BlockQuoteParser.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Markdown\Parsers;
+
+use League\CommonMark\Extension\CommonMark\Node\Block\BlockQuote;
+use League\CommonMark\Node\Node as CommonMarkNode;
+use League\CommonMark\Node\NodeWalker;
+use League\CommonMark\Node\NodeWalkerEvent;
+use phpDocumentor\Guides\MarkupLanguageParser;
+use phpDocumentor\Guides\Nodes\Node;
+use phpDocumentor\Guides\Nodes\QuoteNode;
+use Psr\Log\LoggerInterface;
+use RuntimeException;
+
+use function sprintf;
+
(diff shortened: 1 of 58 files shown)


packages/guides-markdown/src/Markdown/Parsers/BlockQuoteParser.php
```diff
@@ -0,0 +1,62 @@
...
+
+                continue;
+            }
+
+            // leaving the heading node
+            if ($commonMarkNode instanceof BlockQuote) {
+                return new QuoteNode($content);
+            }
+
+            $this->logger->warning(sprintf('BlockQuote CONTEXT: I am leaving a %s node', $commonMarkNode::class));
```
jaapio, line 52:
```suggestion
            $this->logger->debug(sprintf('BlockQuote CONTEXT: I am leaving a %s node', $commonMarkNode::class));
```

packages/guides-markdown/src/Markdown/Parsers/HeaderParser.php
```diff
@@ -0,0 +1,71 @@
...
+            // leaving the heading node
+            if ($commonMarkNode instanceof Heading) {
+                return new TitleNode(
+                    new InlineCompoundNode($content),
+                    $commonMarkNode->getLevel(),
+                    $this->idGenerator->slug($content[0]->toString() ?? '')->lower()->toString(),
+                );
+            }
+
+            $this->logger->warning(sprintf('Header CONTEXT: I am leaving a %s node', $commonMarkNode::class));
```
jaapio, line 61:
```suggestion
            $this->logger->debug(sprintf('Header CONTEXT: I am leaving a %s node', $commonMarkNode::class));
```

packages/guides-markdown/src/Markdown/Parsers/InlineParsers/AbstractInlineTextDecoratorParser.php
```diff
@@ -0,0 +1,80 @@
...
+                if (count($content) === 1 && $content[0] instanceof PlainTextInlineNode) {
+                    return $this->createInlineNode($commonMarkNode, $content[0]->getValue());
+                }
+
+                $this->logger->warning(sprintf('%s CONTEXT: Content of emphasis could not be interpreted: %s', $this->getType(), var_export($content, true)));
+
+                return $this->createInlineNode($commonMarkNode, null);
+            }
+
+            $this->logger->warning(sprintf('%s context does not allow a %s node', $this->getType(), $commonMarkNode::class));
```
jaapio, line 63:
```suggestion
            $this->logger->debug(sprintf('%s context does not allow a %s node', $this->getType(), $commonMarkNode::class));
```

jaapio (commented):
This looks pretty good! Thanks for this amazing amount of work. This seems to set a good foundation of future development to support more markdown features. 

Just some minor changes to finalize this.

packages/guides-markdown/src/Markdown/Parsers/BlockQuoteParser.php
linawolf: I think we should change the message not the logging type. When this happens it means that the commonmarks node was not expected

jaapio (approved):

PR #645, opened 2023-10-30
!!![TASK] Remove all services from render context

jaapio:
The render context should be a service less object only containing the context of the current rendered project/document. This reduces the ways we are generating urls and makes the creation of the object less complex.


packages/guides/src/Handlers/RenderDocumentCommand.php
```diff
@@ -27,6 +27,6 @@ public function getContext(): RenderContext
 
     public function getFileDestination(): string
     {
-        return $this->renderContext->getCurrentFileDestination();
+        return $this->renderContext->getDestinationPath() . '/' . $this->renderContext->getCurrentFileName() . '.' . $this->renderContext->getOutputFormat();
```
linawolf, line 30: I would use sprintf here

linawolf (approved):

PR #650, opened 2023-11-02
[!!!][TASK] Make output an option instead of argument

linawolf:
This way it can be set indepenently from the input

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -35,4 +35,4 @@ jobs:
           dependency-versions: "highest"
 
       - name: "Run guides-cli"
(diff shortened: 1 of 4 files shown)


jaapio (approved):

.github/workflows/documentation.yml
```diff
@@ -35,4 +35,4 @@ jobs:
           dependency-versions: "highest"
 
       - name: "Run guides-cli"
-        run: "vendor/bin/guides -vvv --no-progress docs /tmp/test --fail-on-log"
+        run: "vendor/bin/guides -vvv --no-progress docs --output="/tmp/test" --fail-on-log"
```
jaapio, line 38: you need to use single quotes in the command I think

PR #655, opened 2023-11-04
[BUGFIX] Remove non-existent directives from DI

linawolf:
these have been removed a while back


linawolf:
Merging as a no-brainer

PR #656, opened 2023-11-04
[FEATURE] Enable registration of textrole domains

linawolf:
I need this to register textroles like `:php:interface:`


jaapio:
As explained via slack:
Configure your role like this:

```
phpDocumentor\Guides\RestructuredText\TextRoles\PHPReference:
  tags:
    - { name: 'phpdoc.guides.parser.rst.text_role', 'domain': 'php'}
 ```

PR #659, opened 2023-11-05
[!!!][TASK] Replace the TemplatedNodeRenderer with a configurable Gen…

linawolf:
…eralNoderRenderer

All we need to do it really to match the nodes to their corresponding templates.

We are also able to drop the deprecated TemplateNode

When introducing a directive in an extension and that directive needs a new node it can also register a new renderer for that node.

The matching of nodes to templates can however also be done via DI.

As a proof of concept I added latex support for the confval directive.

This also frees extensions from the need to implement their own noderenderers if all that is needed is passing the node to the template (which is like 95% of nodes introduced.)

This is breaking for any extension using the TemplatedNodeRenderer.


jaapio:
I think I get where you are coming from, but I dislike the current implemented solution. The whole architecture of this library is about nodes. There are multiple ways to add new nodes, via a directive or by adding a new Producer. The compiler does require the node types to perform their actions, and same applies to the renderers. The idea is that there is a one-on-one relationship between the renderers and the nodes. Every node should have it's own renderer. This is the design of the library, changing that violates the architecture, and will make it confusing as there will be multiple ways to render a node. Where it's now just 1. 

If your goal is to make it easier to register a template for a node we can add something in the DI config to make that possible.  So in stead of adding compiler pass per extension we could add a more generic compiler pass that reads a parameter. My suggestion is to create an array like this:

```
    $container->parameters()
        ->set('phpdoc.guides.node_templates', [
            'paragraph_html' => [
                'node' => \phpDocumentor\Guides\Nodes\ParagraphNode::class,
                'template' => 'body/paragraph.html.twig',
                'format' => 'html'
            ]
        ]);
 ```
 
 The compiler pass could read this and add the templated renderer like it does right now.  You can add new templates this way just like we do with the `phpdoc.guides.base_template_paths`.  The advantage is that extensions can change the template in a unified way. And it does not break the way we are doing it now. 
 
Adding new formats will not require extra compiler passes nor an extra extend of a generic renderer. Because you will simply instruct the compiler pass to create more instances of the same templatednoderenderer.

**Review by jaapio — CHANGES_REQUESTED** [review]
I think I get where you are coming from, but I dislike the current implemented solution. The whole architecture of this library is about nodes. There are multiple ways to add new nodes, via a directive or by adding a new Producer. The compiler does require the node types to perform their actions, and same applies to the renderers. The idea is that there is a one-on-one relationship between the renderers and the nodes. Every node should have it's own renderer. This is the design of the library, changing that violates the architecture, and will make it confusing as there will be multiple ways to render a node. Where it's now just 1. 

If your goal is to make it easier to register a template for a node we can add something in the DI config to make that possible.  So in stead of adding compiler pass per extension we could add a more generic compiler pass that reads a parameter. My suggestion is to create an array like this:

```
    $container->parameters()
        ->set('phpdoc.guides.node_templates', [
            'paragraph_html' => [
                'node' => \phpDocumentor\Guides\Nodes\ParagraphNode::class,
                'template' => 'body/paragraph.html.twig',
                'format' => 'html'
            ]
        ]);
 ```
 
 The compiler pass could read this and add the templated renderer like it does right now.  You can add new templates this way just like we do with the `phpdoc.guides.base_template_paths`.  The advantage is that extensions can change the template in a unified way. And it does not break the way we are doing it now. 
 
Adding new formats will not require extra compiler passes nor an extra extend of a generic renderer. Because you will simply instruct the compiler pass to create more instances of the same templatednoderenderer.

linawolf:
It simply makes no sense to introduce a renderer for each node if the renderer is usually doing nothing but passing the node to the template. If more then passing the node to the renderer is needed, ok you can implement a renderer. But it is usually completely unnessesary and just makes it more complicated to introduce custom directives

linawolf:
Also if the node is produced by a compiler pass with the idea I introduce here you could also register the node rendering in the compiler pass

jaapio:
It's the design, each node should have an instance of a node renderer. It's ok to reuse a renderer, but I think we should not break this. I think this new approach makes it less obvious how the rendering is done. Because some nodes will have their own rendering others are sharing a renderer. I don't see how my suggestion is different from yours, you can still do something by configuring an array.  The only thing you need to do in an extension is adding configuration. Your approach just changes the way to have to do the configuration. But introduces a violation to the original design. 

Besides that with the current implementation I can overwrite the rendering by registering a new service with the same ID, which is not possible with your implementation. As node renderers to not have a priority. 

Sorry, this is really a no go for me.

linawolf:
I think having the infomation in all these different places makes it unnessesary complicated to add a directive.

PR #661, opened 2023-11-06
[TASK] Add failing tests for non-removed class node

linawolf:
@jaapio could you have a look maybe? It seems like even though the  ClassNodeTransformer returns null the node does not get properly removed somewhere in the shadow trees. And I dont really get what the problem is.


jaapio:
Thanks for reporting, this is fixed now.

PR #665, opened 2023-11-07
[TASK] bump league uri depenciency

jaapio:
allow 6.5 and 7.0


jaapio:
@thommyhh here you go :-)

PR #666, opened 2023-11-08
[TASK] Validate guides.xml files against scheme

linawolf:


packages/guides-cli/resources/schema/guides.xsd
```diff
@@ -2,12 +2,13 @@
 <xsd:schema
     targetNamespace="https://www.phpdoc.org/guides"
     xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+    xmlns="https://www.phpdoc.org/guides"
```
jaapio, line 5: we should publish the xsd

linawolf: First we should make it work?

linawolf:
@jaapio all tests are green now, please review

packages/guides-cli/resources/schema/guides.xsd
wouterj: I wouldn't publish the XSD on an uri (but instead reference a local file like done in the examples in this PR). You need proper version management when publishing the XSD's, something you get from free when you recommend people to use the XSD installed in their vendor directory.

(XML namespaces being a "uri" has no relation to the web)

jaapio (approved):

PR #669, opened 2023-11-09
[FIX] support single document rendering

jaapio:
To make it possible to extend the html rendering with a single document we should allow customization of the html rending stack. The easiest way to do this this to add a new renderer that is reusing the delegating renderer from the html part. By allowing extensions to define their own delegating renderer with a factory set, we can switch towards another stack.

This is a rather complex use case, which we do not support out of the box but needed for typo3.


linawolf:
Why don't we support rendering the whole document in one file? Others might find it helpful too (very helpful for printing or exporting) and we could have integration tests for this to make sure it is working with update?

linawolf (approved):

PR #673, opened 2023-11-12
[FEATURE] Use PSR-20 time and attach time to projectNode

linawolf:


tests/Integration/IntegrationTest.php
```diff
@@ -64,9 +66,13 @@ public function testHtmlIntegration(
             system('mkdir ' . escapeshellarg($outputPath));
 
             $this->prepareContainer($configurationFile);
+
             $command = $this->getContainer()->get(Run::class);
+
             assert($command instanceof Run);
 
+            $command->setClock(new MockClock('2023-01-01 12:00:00'));
```
jaapio, line 74: You can set this in the `\phpDocumentor\Guides\DependencyInjection\TestExtension` then there is no reason the have the setClock method. You can register the MockClock there as a service with a certain argument.

tests/Integration/IntegrationTest.php
```diff
@@ -39,6 +40,7 @@ class IntegrationTest extends ApplicationTestCase
     protected function setUp(): void
     {
         setlocale(LC_ALL, 'en_US.utf8');
+        // todo: mock time
```
jaapio, line 43: This seems to be obsolete?

tests/Integration/IntegrationTest.php
linawolf: How could I set the mock time there?

linawolf: Ok, figured it out, setClock is removed

linawolf:
@jaapio Please have another look

jaapio (approved):

PR #680, opened 2023-11-16
[DOCS] Clarify parameter usage for the "guides" binary]

garvinhicking:
Adjusts the meaning of mainly input-file to be clearer.

Adjusts a wrong description of the "output" option.

Enclose arguments with quotes to make it clearer, that these are
the actual values.

Also unifies not having a "." at the end of sentences.


linawolf (approved):
Thanks a lot for your contribution! I think this makes the commands clearer to understand!

PR #686, opened 2023-11-18
[TASK] Replace Integration tests with xmllint

linawolf:
Linting xml files is not really the responsibility of integration tests.

As a plus using xmllint works on all guides.xml including those in our doc directly and the basic one


linawolf:
@jaapio any idea why tools/xmllint.sh is not found in the github actions?

jaapio:
You need to add a the checkout step like the other. Jobs have.

jaapio (approved):

PR #687, opened 2023-11-19
[TASK] Split integration tests in content and full test

linawolf:
This way changes to the general layouts do not change all the tests.

As a bonus with php tools/integration-test-copy-baseline.php you can update the baseline of your tests if still many tests should be affected


jaapio (approved):
Please rebase, so we can merge this.

Issue #688, opened 2023-11-20
"emphasize-lines" in "code-block" directive not considered


brotkrueml:
In the following example:

```rst
..  code-block:: python
    :emphasize-lines: 3,5

    def some_function():
        interesting = False
        print('This line is highlighted.')
        print('This one is not...')
        print('...but this one is.')
```

does not emphasize the lines.

In the CodeNode there is also no property for this:

```
object(phpDocumentor\Guides\Nodes\CodeNode)#2467 (6) {
  ["classes":protected]=>
  array(0) {
  }
  ["options":protected]=>
  array(0) {
  }
  ["value":protected]=>
  string(148) "def some_function():
    interesting = False
    print('This line is highlighted.')
    print('This one is not...')
    print('...but this one is.')"
  ["startingLineNumber":"phpDocumentor\Guides\Nodes\CodeNode":private]=>
  NULL
  ["caption":"phpDocumentor\Guides\Nodes\CodeNode":private]=>
  NULL
  ["language":protected]=>
  string(6) "python"
}
```

PR #690, opened 2023-11-20
[FEATURE] Support emphasize-lines in code-block directive

linawolf:
resolves https://github.com/phpDocumentor/guides/issues/688

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Directives/CodeBlockDirective.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Directives/CodeBlockDirective.php
@@ -10,7 +10,9 @@ use phpDocumentor\Guides\RestructuredText\Directives\OptionMapper\CodeNodeOption
 use phpDocumentor\Guides\RestructuredText\Parser\BlockContext;
 use phpDocumentor\Guides\RestructuredText\Parser\Directive;
 use phpDocumentor\Guides\RestructuredText\Parser\DirectiveOption;
+use Psr\Log\LoggerInterface;
 
+use function preg_match;
 use function trim;
 
 /**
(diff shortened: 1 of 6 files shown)


packages/guides-restructured-text/src/RestructuredText/Directives/CodeBlockDirective.php
```diff
@@ -101,4 +102,15 @@ private function setCaptionBasedOnOptions(array $options, CodeNode $node): void
 
         $node->setCaption($caption);
     }
+
+    /** @param DirectiveOption[] $options */
+    private function setEmphasizeLinesBasedOnOptions(array $options, CodeNode $node): void
+    {
+        $emphasizeLines = null;
+        if (isset($options['emphasize-lines'])) {
+            $emphasizeLines = (string) $options['emphasize-lines']->getValue();
```
jaapio, line 111: maybe we should validate the input?

jaapio (commented):
Just a small suggestion, rather then that this looks fine

packages/guides-restructured-text/src/RestructuredText/Directives/CodeBlockDirective.php
linawolf: how would you validate it? regular expression?

jaapio: yes, that would do the trick, then log a warning when it is invalid.

linawolf: added a regex

jaapio (approved):

PR #699, opened 2023-11-21
[TASK] Add "release" and "copyright" as valid "<project>" xsd

garvinhicking:
This change in the XSD will allow to utilize "release" and "copyright" attributes from the old Sphinx Settings.cfg


garvinhicking:
(Will be superseded by Linas Patch)

PR #700, opened 2023-11-21
[FEATURE] Introduce release and copyright information on project

linawolf:
In sphinx the version is the minor version i.e. 12.4 while the release is considered as exact version i.e. 12.4.9-beta or some-such.

Additionally copyright information can be stored project-wide.


garvinhicking (approved):
Great!! :-)

jaapio:
I think this PR on it's own is ok, but I'm wondering where it will end, we are adding more properties to the project node, but I think we would need some different way to add this. I know that you will also need other settings so maybe this is the time to try to implement some more generic way to store this. 

We should keep it simple for now, what if we would add a key value store on the project to add this kind of settings?

linawolf:
The other settings, those that do not exist in sphinx, we added within our theme extension. For example the edit on github button links etc

PR #705, opened 2023-11-22
[TASK] Increases verbose output when debugging (-vvv) is set

garvinhicking:
To debug https://github.com/TYPO3-Documentation/render-guides/issues/83 we need more verbose output what directories/files are used.

Sadly, `$output` isn't available within this file (yet) and I don't know how to properly detect the "-vvv" argument at this point. This would be vital before merging this PR.

### DIFF
--- a/packages/guides-cli/bin/guides
+++ b/packages/guides-cli/bin/guides
@@ -34,15 +34,23 @@ $input = new ArgvInput();
 
 $containerFactory = new ContainerFactory([new ApplicationExtension()]);
 
+// TODO: How to use $output at this point for coloured output and access
+//       $output->isVerbose() / $output->isDebug()?
 if (is_file($vendorDir . '/../guides.xml')) {
+    echo "DEBUG: Using vendorDir guides.xml file ($vendorDir/../guides.xml)\n";
     // vendor folder was placed directly into the project directory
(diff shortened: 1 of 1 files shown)


garvinhicking:
https://github.com/TYPO3-Documentation/render-guides/pull/119 Is needed for this, too.

garvinhicking:
One more problem with this.

The github action calls the binary like this:

````
- uses: TYPO3-Documentation/render-guides@main
      id: render-guides
      if: steps.enable_guides.outputs.GUIDES == 'true'
      with:
        working-directory: t3docsproject
        config: ./Documentation/
        output: ./RenderedDocumentation/Result/project/0.0.0
```

but we can't put `vvv: something` nor `verbose: vvv` into the action, because it's a short-hand command.

How could we then access verbose output when being called within the github action? Do we need to "invent" a new `guides`  parameter like `--debugging=yes` here, so that we could use a github action like this:

````
- uses: TYPO3-Documentation/render-guides@main
      id: render-guides
      if: steps.enable_guides.outputs.GUIDES == 'true'
      with:
        working-directory: t3docsproject
        config: ./Documentation/
        debugging: yes
        output: ./RenderedDocumentation/Result/project/0.0.0
```

?

linawolf:
Setting this on draft so it is not accidently merged while still containing echo

PR #710, opened 2023-11-23
[TASK] Improve warnings and debug info when references are not found.

linawolf:
Also apply spaceless to the link and reference templates.

Suggest probable targets by Levenstein distance


tests/Integration/tests/body-elements/line-blocks/expected/index.html
```diff
@@ -1,38 +1,17 @@
...
-                    are preserved.
-                </div>
-            </div>
-            <div class="line">
-                Continuation lines are wrapped
-                portions of long lines; they begin
-                with spaces in place of vertical bars.
-            </div>
-        </div>
+            <p>| Line blocks are useful for addresses,
```
garvinhicking, line 5: Shouldn't this remain using `<div>` output style tags?

garvinhicking (commented):
I love the extra debugging 🥇 
Can't comment on Levenshtein, I don't know him well. ;)

tests/Integration/tests/body-elements/line-blocks/expected/index.html
linawolf: arg true. Copying real output into expected should exclude unrelated skipped tests

packages/guides/src/ReferenceResolvers/DocReferenceResolver.php
```diff
@@ -29,11 +34,27 @@ public function resolve(LinkInlineNode $node, RenderContext $renderContext): boo
...
+                sprintf(
+                    'Document with name "%s" not found, required in file "%s". %s',
+                    $canonicalDocumentName,
+                    $renderContext->getCurrentFileName(),
+                    $didYouMean,
+                ),
+                array_merge($renderContext->getLoggerInformation(), $node->getDebugInformation()),
+            );
+
+            return true;
```
jaapio, line 57: Should this not be false? As there is nothing resolved. 

Maybe we should move the suggestions to a different resolver with the lowest priority, So the chain would be:

- DocReference
- Interlink
- ....
- SuggestionDocs
- SuggestionInterlink

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
```diff
@@ -27,18 +33,55 @@ public function resolve(LinkInlineNode $node, RenderContext $renderContext): boo
...
+            $this->logger->warning(
+                sprintf(
+                    'Inventory with name "%s" could not be resolved in file "%s". ',
+                    $domain,
+                    $renderContext->getCurrentFileName(),
+                ),
+                array_merge($renderContext->getLoggerInformation(), $node->getDebugInformation()),
+            );
+
+            return true;
```
jaapio, line 45: Same here, should this not be false?

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
```diff
@@ -27,18 +33,55 @@ public function resolve(LinkInlineNode $node, RenderContext $renderContext): boo
...
+                sprintf(
+                    'Inventory with name "%s" does not contain group %s, required in file "%s". ',
+                    $domain,
+                    $group,
+                    $renderContext->getCurrentFileName(),
+                ),
+                array_merge($renderContext->getLoggerInformation(), $node->getDebugInformation()),
+            );
+
+            return true;
```
jaapio, line 61: Same here, should this not be false?

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
```diff
@@ -27,18 +33,55 @@ public function resolve(LinkInlineNode $node, RenderContext $renderContext): boo
...
+                    $domain,
+                    $target,
+                    $group,
+                    $renderContext->getCurrentFileName(),
+                    $didYouMean,
+                ),
+                array_merge($renderContext->getLoggerInformation(), $node->getDebugInformation()),
+            );
+
+            return true;
```
jaapio, line 84: Same here, should this not be false?

jaapio (commented):
I see that you repeated the suggestion algorithm multiple times. It feels a bit strange to have it spread over the code as it is part of our resolver logic. 

A suggestion to move it to a separate resolver, which is aware of multiple strategies? or maybe multiple separate resolvers? 

I think the changes of `return false` to `return true` might cause issues, so please let's talk about that why it's required to have this change.

packages/guides/src/ReferenceResolvers/DocReferenceResolver.php
linawolf: If we would do that we would output multiple warnings for the same reference?

linawolf:
Agreed with @jaapio that this PR needs more though...

linawolf:
Gonna split this up into smaller, reviewable parts and apply suggestions

PR #714, opened 2023-11-24
[TASK] add debug level echo for config

jaapio:
Enviroment variable SHELL_VERBOSITY can be used or `-vvv` to echo the used config files


jaapio:
supersedes #705

PR #717, opened 2023-11-25
[TASK] Move all interlink tests into one directory

linawolf:


linawolf:
Merging this as a no-brainer

PR #720, opened 2023-11-25
[TASK] Improve reference warnings

linawolf:
Scaled down version of the improved warnings. The suggestion and the propper storing and handling of warnings to prevent duplicates will come in a follow up


packages/guides/src/Interlink/InventoryLink.php
```diff
@@ -38,4 +39,14 @@ public function getTitle(): string
     {
         return $this->title;
     }
+
+    public function getOriginalKey(): string
```
jaapio, line 43: Why do we need this? It seems to be unused right now, if it's needed for a follow up please remove the methods for now, so we do not have dead code.

jaapio (approved):

packages/guides/src/Interlink/InventoryLink.php
linawolf: This was a leftover from splitting up the commits, removed it

PR #723, opened 2023-11-26
[BUGFIX] Allow empty lines between field list items

linawolf:
They are also allowed in Sphinx and we frequently have them in TYPO3.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldListRule.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldListRule.php
@@ -55,6 +55,14 @@ final class FieldListRule implements Rule
         while ($iterator->valid() && $this->isFieldLine($iterator->current())) {
             $fieldListItemNodes[] = $this->createListItem($blockContext);
             $iterator->next();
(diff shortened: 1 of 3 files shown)


packages/guides-restructured-text/src/RestructuredText/Parser/Productions/FieldListRule.php
```diff
@@ -55,8 +55,15 @@ public function apply(BlockContext $blockContext, CompoundNode|null $on = null):
         while ($iterator->valid() && $this->isFieldLine($iterator->current())) {
             $fieldListItemNodes[] = $this->createListItem($blockContext);
             $iterator->next();
+            while ($iterator->valid() && LinesIterator::isEmptyLine($iterator->current())) {
```
jaapio, line 58: If you use peek here to get the next line we can remove the prev call... I would rather see prev on the line iterator as a hack. So if we can avoid usage we should do this. 

As far as I know there is right now 1 location using prev, which is on my list to refactor.

jaapio (changes requested):
I would like to see if we can remove prev call by refactoring the foreach to skip the empty lines. 

Please let me know when you need help

jaapio (approved):

PR #744, opened 2023-12-09
[BUGFIX] Rename textrole template

linawolf:
"unknown" hat the second n missing, thanks to @brotkrueml for pointing this out


linawolf:
Merging as non-brainer

PR #755, opened 2023-12-14
[TASK] Improve keys of bootstrap tabs

linawolf:


linawolf:
Merging as low-brainer only affecting bootstrap package

PR #756, opened 2023-12-14
[BUGFIX] Correct aria-labels for tabs

linawolf:


linawolf:
Merging as no-brainer

PR #758, opened 2023-12-16
Move section anchors before the section title

wouterj:
Hyperlink targets (or Anchor nodes) of a section are currently parsed as child nodes. This is alright, but causes a rendering issue as the anchors are placed below the title node. This means that referencing such anchors scrolls the section title outside of the view.

I don't like my solution in this PR, but it at least fixes the bug by placing the anchor nodes before the title node (which is how Sphinx does it). If anyone has a better suggestion, please tell me :)

| *Before* | *After*
| --- | ---
| image | image

### DIFF
--- a/packages/guides/resources/template/html/structure/section.html.twig
+++ b/packages/guides/resources/template/html/structure/section.html.twig
@@ -1,5 +1,4 @@
 <div class="section{% if node.classes %} {{ node.classesString }}{% endif %}" id="{{ node.title.id }}">
-    {{ renderNode(node.title) }}
     {% for childNode in node.children %}
         {{ renderNode(childNode) }}
     {% endfor %}
--- a/packages/guides/resources/template/tex/structure/section.tex.twig
+++ b/packages/guides/resources/template/tex/structure/section.tex.twig
@@ -1,5 +1,3 @@
-{{ renderNode(node.title) }}
-
 {% for childNode in node.children %}
     {{ renderNode(childNode) }}
 {% endfor %}
--- a/packages/guides/src/Nodes/SectionNode.php
+++ b/packages/guides/src/Nodes/SectionNode.php
@@ -13,7 +13,7 @@ final class SectionNode extends CompoundNode implements LinkTargetNode
 
     public function __construct(private readonly TitleNode $title)
     {
-        parent::__construct();
+        parent::__construct([$title]);
     }
 
     public function getTitle(): TitleNode
--- a/packages/guides/tests/unit/Compiler/NodeTransformers/MoveAnchorTransformerTest.php
+++ b/packages/guides/tests/unit/Compiler/NodeTransformers/MoveAnchorTransformerTest.php
@@ -50,7 +50,7 @@ final class MoveAnchorTransformerTest extends TestCase
         $this->documentNodeTraverser->traverse($document, $context);
 
         self::assertCount(1, $context->getDocumentNode()->getChildren());
-        self::assertCount(1, $section->getChildren());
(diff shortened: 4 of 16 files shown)


linawolf:
I believe this is something that should be solved in the Node-Transformers during compilation, not the template... If you want to I can make a proposal

wouterj:
The issue is that the title node is not part of the nodes children, but rendered independently. Maybe we can also make it part of the children and fix the order in a node transformer?

linawolf:
Shouldnt the ancors be displayed before the actual section starts?

linawolf:
we could do something like sectionNode-> getAnchors and then remove the anchors from the general children?

jaapio:
In phpDocumentor we use twig functions to filter the objects. What might help here is a filter option to pass an expression to the node filter nodes?

wouterj:
@jaapio the issue is that we do not want filter as we only want the preceding anchor nodes (not the ones that are located at a specific paragraph in the section). PHP doesn't have nice `takeWhile()` or `dropWhile()` functions.

@linawolf Sphinx renders them in the section. Let me try your suggestion of adding a dedicated `anchorNodes` property.

wouterj:
I think I found a much easier (and more correct) solution. The `MoveAnchorTransformer` already pushes the anchors in front of a section as the first child of that section. However, we weren't treating the section's `TitleNode` as a node in the tree.

By adding the `TitleNode` to the section children, order is automatically fixed and we can simplify the `section.*.twig` template as it doesn't need to take care of rendering the title separately. This is however a small BC break for themes overriding this template.

I also feel like treating the `TitleNode` as a child of the section is more correct.

linawolf:
The title is not a child of the section really but a title defines the section itself. I wouldn't want different Transformers to start messing here with the order, really. And also there are situations in which it makes sense to treat the title independently from the content, really.

wouterj:
I think in general, we don't want transformers mess with the order of any node unless it's part of a domain rule (e.g. with anchors before a section or a class directive without content). I don't think this specifically applies to titles.
There is still a possibility to treat the title independent from the content, using `TitleNode#getTitle()` this is used in the PHP code (e.g. when dealing with TOCs I believe).

When implementing the first suggestion, have a `SectionNode#anchorNodes` property, there were a couple things I didn't like about the code: Generic things like the move anchor transformer had to specifically change its behavior when dealing with section nodes. The same applies to the collect references transformer, as the section anchors were no longer part of the section children.
Moreover, I felt like this coupled the compiling phase (where position doesn't matter) way to much to the rendering phase (were position matters).

FWIW, docutils also treats the title node as a child of the section node: https://github.com/docutils/docutils/blob/f725d4a2742f997a95b9de53553642044df475ab/docutils/docutils/parsers/rst/states.py#L379-L387

linawolf (approved):
I  gave it another look and I think it would work like that. (Breaking change to the templates but we still haven't releases 1.0 yet). If we wanted to style the header we could also do that in the headers template. so no need really to output it differently int the sections template. 

And I can see you also didnt break the link to header functionality, so We (TYPO3) would be ok with this change. But lets @jaapio also have a look

jaapio:
I think this change is great. However I would like to see the `getTitle` method to use the children collection as well. Because right now we have 2 references and as the nodes are immutable (mostly) I'm afraid that the dubble administration will cause issues. 
For performance reasons we could use a weakreference to cache the title node locally in the node. That will be gone once we change the title node?

linawolf:
@jaapio can we do that in a follow up or must it be solved here?

jaapio:
I will address this.

PR #759, opened 2023-12-16
[BUGFIX] Correctly mark active and current menu states

linawolf:


packages/guides/src/Compiler/Passes/GlobalMenuPass.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public function run(array $documents, CompilerContext $compilerContext): array
+    {
+        $projectNode = $compilerContext->getProjectNode();
+        try {
+            $rootDocumentEntry = $projectNode->getRootDocumentEntry();
+        } catch (Throwable) {
+            // Todo: Functional tests have not root document entry
+            return [];
+        }
```
jaapio, line 33: To be addressed in a follow up, I think it makes sense to remove the exception that is been thrown by `getRootDocumentEntry` if this is a common pattern. So we use exceptions just for exception.

packages/guides-theme-bootstrap/resources/template/body/menu/navbar/menu-level.html.twig
```diff
@@ -3,8 +3,8 @@
     {% apply spaceless %}
         <li class="nav-item">
             <a href="{{ renderLink(entry.url) }}"
-               class="nav-link {% if entry.options.active %} active {% endif %}"
-                    {% if entry.options.active %}  aria-current="page" {% endif %} >
+               class="nav-link  {%- if menu.currentPath == entry.url %} current {%- endif -%}{% if entry.url in menu.rootlinePaths %} active {%- endif -%}"
+                    {%- if menu.currentPath == entry.url %} aria-current="page" {% endif %} >
                 {{ renderNode(entry.value.value) }}
             </a>
```
jaapio, line 9: Maybe we should extract this to a menu item link template? to remove the duplication in the templates.

packages/guides/src/RenderContext.php
```diff
@@ -152,6 +152,24 @@ public function getCurrentFileName(): string
...
+
+        $rootline = [];
+        $documentEntry = $this->getCurrentDocumentEntry();
+        $rootline[] = $documentEntry->getFile();
+        while ($documentEntry->getParent() instanceof DocumentEntryNode) {
+            $documentEntry = $documentEntry->getParent();
+            $rootline[] = $documentEntry->getFile();
+        }
+
+        return $rootline;
```
jaapio, line 170: As the render context is created per document, It would make sense to cache this rootline in context. That would reduce the impact of calling this method. Or should we store the rootline in the document entry?

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace phpDocumentor\Guides\Compiler\Passes;
+
+use phpDocumentor\Guides\Compiler\CompilerContext;
+use phpDocumentor\Guides\Compiler\CompilerPass;
+use phpDocumentor\Guides\Nodes\DocumentNode;
+use phpDocumentor\Guides\Nodes\Menu\NavMenuNode;
+use Throwable;
+
+class GlobalMenuPass implements CompilerPass
```
jaapio, line 13: Please mark this class as final.

jaapio (approved):
I see some possible improvements. But as we talked about on slack you have some follow up changes ready. 

Nothing blocking to I will merge this. Thanks!

PR #760, opened 2023-12-16
Fix vendor directory discovery if it is more than 1 level deep

wouterj:
I've been applying this patch consistently the past year in my test project. Due to the submodule/symlink structure of the project, the vendor directory was 4 levels deep. The while loop actually didn't work on anything other than a `vendor` directory in the `guides-cli/bin` directory.


linawolf (approved):

linawolf:
Thanks for this fix

PR #761, opened 2023-12-16
Fix using < or > in a reference name

wouterj:
These characters are valid and only the last `<...>` must be treated as the embedded URL.

This PR removes using the inline lexer inside references, reSt doesn't allow nested inline nodes and the only thing required is getting the embedded URL (if there is any). This reduces a lot of complexity

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/InlineLexer.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/InlineLexer.php
@@ -23,8 +23,6 @@ final class InlineLexer extends AbstractLexer
     public const ANONYMOUS_END = 3;
     public const LITERAL = 5;
     public const BACKTICK = 6;
-    public const EMBEDED_URL_START = 9;
-    public const EMBEDED_URL_END = 10;
     public const NAMED_REFERENCE = 11;
     public const ANONYMOUSE_REFERENCE = 12;
     public const COLON = 13;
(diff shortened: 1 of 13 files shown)


packages/guides-restructured-text/src/RestructuredText/Parser/InlineLexer.php
```diff
@@ -60,10 +60,9 @@ protected function getCatchablePatterns(): array
             '(?<=^|\s)[a-z0-9-]+_{2}', //Inline href.
             '(?<=^|\s)[a-z0-9-]+_{1}(?=[\s\.+]|$)', //Inline href.
             '``.+?``(?!`)',
+            '<[^<]+?>(?=`)',
```
linawolf, line 63: This will catch the same pattern within normal texts. The normal text will then be treated as one token missing other things that might be inside. What will for example happen to the following: 

:code:`<` and then :code:`>` Will this give us a false positive? It might. At least add tests for it?

linawolf (changes requested):
There might be issues with false positives here

wouterj:
You are correct, this was too fragile. I've now rewritten the code to delegate parsing the embedded URI to the rules/roles that need it and removing it from the global inline lexer. This is the safest and shouldn't break anything, given embedded URIs do not exist on their own.

linawolf:
I like it, it will also make things easier for us in the php-domain text-roles

linawolf (approved):

jaapio (approved):

PR #762, opened 2023-12-17
Upgrade to latest league/csv

greg0ire:
That version adds more insight on what is returned by `League\Csv::getRecords()`, which is now documented as an `Iterator<array<mixed>>`. Since we are using strict types, we are already assuming that mixed will in fact always be a string or null. Let's make the assumption clear to static analysis tools.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -401,35 +401,36 @@
         },
         {
             "name": "league/csv",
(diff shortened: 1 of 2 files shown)


jaapio (approved):

jaapio:
Thanks!

PR #763, opened 2023-12-18
Guides Code: Small improvements

wouterj:
While testing the code extension in the Symfony Docs build, I found 2 improvements:

- Language names in the XML config where "normalized" (i.e. dashes are replaced with underscores). We should not change language names (Symfony uses e.g. `php-attribute` to denote a PHP example with attribute config).
- The `Highlighter` class was final and there was no interface. This means I couldn't decorate it to add some weird custom Symfony docs logic. Nor does not allow people to plug in other server-side highlighters. I've added an interface.

Also, we need a subsplit and Composer package for this new extension @jaapio


brotkrueml:
Just a small note: I build an independent Twig filter which we use for the TYPO3 guides rendering, maybe this is helpful ;-)
https://github.com/brotkrueml/twig-codehighlight
For now, there is no version published and no Packagist package available, but will come soon.

jaapio (approved):

jaapio:
Here you go: https://packagist.org/packages/phpdocumentor/guides-code

wouterj:
Thank you! We're getting extremely close to a green, no-skip, testsuite for Symfony :)

PR #765, opened 2023-12-18
[BUGFIX] Treat terms with textroles as definition list

linawolf:
if they are no field list


packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DefinitionListRule.php
```diff
@@ -156,8 +156,7 @@ private function isDefinitionTerm(string|null $currentLine, string|null $nextLin
             return false;
         }
 
-        // This a field list
-        if (str_starts_with(trim($currentLine), ':')) {
+        if (FieldListRule::isFieldLine($currentLine)) {
```
jaapio, line 159: I understand why you are sharing this code, but I think it's wise to keep the Productions decoupled. In this case I would go for a duplication of the check, as it is very specific for this rule.

jaapio (changes requested):

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/DefinitionListRule.php
linawolf: I think the code is not specific to the Definition List. What we do here is: If we encounter a FieldListline within what we think is a definition list we stop the definition list so that the field list can take over from here on. So the rule really belongs to the Field list

linawolf: But it is cleaner to call applies on the field list probably?

linawolf: Nevermind, didnt work out as applies expects a blockcontext

jaapio (approved):

PR #770, opened 2023-12-19
Render inline nodes in sidebar title

wouterj:
Sidebar titles are allowed to contain inline markup.

This is a small BC break, as we change the property of the section node.


linawolf (approved):
Good catch

PR #771, opened 2023-12-19
[TASK] Add skipped test for escaping in references

linawolf:


linawolf:
@wouterj I believe this didn't work even before your change. Didn't think of this before. But it should be possible to escape in rst

wouterj:
Good call, thanks for adding a test!

I've made a PR to the branch of this PR in https://github.com/phpDocumentor/guides/pull/772 There are 2 minor syntax issues. Still, not everything is working, but the references at least are :relieved:

jaapio (approved):

PR #772, opened 2023-12-19
Fix usage of ref text role

wouterj:
A ref role can only reference anchors, not implement references. Besides, the `_` is not part of the anchor name, but part of the syntax to define an anchor.

Fixing the syntax in the test case shows that references work as expected using escape characters, but there is an issue in the rendering of the section title in the reference when using named phrases. I can look into this next week, if needed.


jaapio:
Hm, this is closed but it should not, could you please change the base branch so we can merge this.

wouterj:
see https://github.com/phpDocumentor/guides/pull/779

PR #776, opened 2023-12-22
[POC] Cachable inline rule match

jaapio:
This is an attempt to cut down the parser duration. This change already stripped of 300ms on a total of 5.6 seconds on the symfony docs on just the parsing. (~6%) 

I do not really like the current implementation of the cache so I would like to get some feedback from you.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/InlineParser.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/InlineParser.php
@@ -8,6 +8,7 @@ use Exception;
 use phpDocumentor\Guides\Nodes\Inline\PlainTextInlineNode;
 use phpDocumentor\Guides\Nodes\InlineCompoundNode;
 use phpDocumentor\Guides\RestructuredText\Parser\Productions\InlineRules\InlineRule;
+use phpDocumentor\Guides\RestructuredText\Parser\Productions\InlineRules\MatchCachable;
 
 use function usort;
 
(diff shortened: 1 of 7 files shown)


jaapio:
@wouterj @greg0ire could you please have a look, and maybe you have some idea's how to continue here. If I remember it correctly symfony has some equal cache logic in it's core.

packages/guides-restructured-text/src/RestructuredText/Parser/InlineParser.php
```diff
@@ -8,34 +8,53 @@
...
+
+            foreach ($this->rules as $key => $inlineRule) {
                 $node = null;
-                if ($inlineRule->applies($lexer)) {
-                    $node = $inlineRule->apply($blockContext, $lexer);
+                if ($inlineRule->applies($this->lexer)) {
+                    if ($inlineRule instanceof MatchCachable && $inlineRule->isCacheable()) {
+                        $this->cachedRules[$this->lexer->token->type] = $inlineRule;
+                        unset($this->rules[$key]);
+                    }
```
jaapio, line 55: Because we can remove the cached rules from the rules array we can remove about 1M calls to the parser rules. The inline rules are matched very often as very word and sometimes char can be a token. This foreach takes about 30% of the parsing time. 

image

wouterj:
See my description in #784 (which does something similar in another place). The current `supports()` method is too generic: implementations might check other things than the type to determine "support". This means we cannot cache token type to inline rule.

As this library is not 1.0 yet, I would advise taking the simple path of changing the `support()` contract so that implementations are forced to do their check only on token type (unless we have a use-case).

jaapio:
I think it's more likely that the parser needs more complex logic to determine the correct strategy of handling a token, right now we are not using lookahead very much, but this would be possible. 
So that's why I chose this path for the parser stategies. 
Rendering is much easier to I do agree with the suggested approach there.

PR #778, opened 2023-12-24
[TASK] Introduce Interfaces for Inventory loading

linawolf:
So that they can be exchanged by DI, for example to load inventories from composer


linawolf:
Merging this as a low-brainer

PR #780, opened 2023-12-24
Do not add anchor to URL for document (root) titles

wouterj:
We already had a flag for menu entries to mark them as "document root", but it was always set to `false`. This PR sets it to `true` when then menu entry references a document root title (in contrast to a sub title within that document).

I've also updated the renderer to not add the anchor when it's a root title. This makes the UX nicer, as you don't need to jump to the document root when you're just navigating to the document start.

### DIFF
--- a/packages/guides/src/Compiler/NodeTransformers/MenuNodeAddEntryTransformer.php
+++ b/packages/guides/src/Compiler/NodeTransformers/MenuNodeAddEntryTransformer.php
@@ -73,7 +73,7 @@ class MenuNodeAddEntryTransformer implements NodeTransformer
                     $documentEntry->getFile(),
                     $documentEntry->getTitle(),
                     [],
-                    false,
+                    true,
                     1,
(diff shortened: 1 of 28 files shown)


linawolf:
I am skeptical about this. Linking to a HTML page instead of the title only works in single rendering mode. Not when all HTML output is rendered in a single page. 

Also at least in TYPO3 we treat sections as independent and they can and are moved around. So we frequently link to them by anchor.

While I do agree to appropriately setting the flag I would not remove the anchors. If for some reason a document does not want to render them, that can be decided in the template. However we do need anchors on a documents main section and use them frequently

linawolf:
With https://github.com/phpDocumentor/guides/pull/799 Menu entries for pages and sections are different Node types. You should therefore be able to define different templates for them.

linawolf:
@wouterj closing this, you should be able to solve it with the templates now?

PR #783, opened 2023-12-26
[FEATURE] Handle Global Menu depth independently from toc

linawolf:
If the toctree on the root page is limited to a certain level count, like `:maxdepth: 1` still the complete page tree is displayed in Sphinx.

However as not ever project needs a full document tree and the creation of the document tree is time-consuming, an option for the guides.xml, max-menu-depth, was introduced. `max-menu-depth = 0` disables the global menu all together.


jaapio:
Suggestion: It is a bit odd to me to set the max-depth to 0 to disable the global menu. I would rather see `0` as unlimited which is the default. if you want to have it different you can increase it to 100. 

Other way of disabling the menu:

Disable the menu could be a different option. however from a user point of view, people could also remove the compilerpass when they setup their application. Which might be an option you could provide in your typo3 app? On create of the DI container we know the settings so we could remove the `GlobalMenuPass` from the collected list of compiler passes. This way the compiler pass doesn't have to disable itself. 

I'm open for a discussion about my suggestions.

linawolf:
@jaapio I applied your suggestions. We dont need to disable this so whoever wants to disable it probably can remove the pass like you said or introduce the option

jaapio (approved):

PR #784, opened 2023-12-27
Cache matching node renderers

wouterj:
By caching which node renderer can be used for which node, we can avoid a lot of foreach loops.

From a quick local test (with pcov enabled), this improved run time by 50%. Blackfire diff: https://blackfire.io/profiles/compare/9aa42af9-7692-418c-9d34-f7e010a22f8f/graph

We can optionally do this in a non-BC breaking way by introducing a new cacheable interface with a `supportsFqcn()` method (similar to Symfony's [voter caching](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Security/Core/Authorization/Voter/CacheableVoterInterface.php)). However, I can find no node renderer in guides and Symfony that does something other than `instanceof`. So I would favor simplicity for the years to come over this BC break.
But I'm happy to discuss rewriting this to a smooth upgrade path if this helps TYPO3.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/NodeRenderers/Html/AdmonitionNodeRenderer.php
+++ b/packages/guides-restructured-text/src/RestructuredText/NodeRenderers/Html/AdmonitionNodeRenderer.php
@@ -21,6 +21,7 @@ use phpDocumentor\Guides\RestructuredText\Nodes\AdmonitionNode;
 use phpDocumentor\Guides\TemplateRenderer;
 
 use function implode;
+use function is_a;
 
 /** @implements NodeRenderer<AdmonitionNode> */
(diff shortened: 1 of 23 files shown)


jaapio:
I would suggest to take the reverse path, do it this way... and if we need a separate renderer for nodes by content we can always introduce this in a different way especially because this is a huge performance improvement. I think right now nobody is using the generic supports method right now. 

Thanks for this improvement. Wouter!

linawolf (approved):
Thanks for this great improvement. The guides are already so fast and still so much opportunity to improve, amazing!

I don't think we have any custom node renderers in TYPO3 so for us it is not breaking.

Great work!

PR #790, opened 2023-12-29
[FEATURE] Support Menu entries with overriden titles and external urls

linawolf:
This change is breaking if someone extended the menu nodes or build their own ones

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Toc/ToctreeBuilder.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Toc/ToctreeBuilder.php
@@ -4,46 +4,49 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Guides\RestructuredText\Toc;
 
-use phpDocumentor\Guides\Nodes\Menu\ParsedMenuEntryNode;
+use phpDocumentor\Guides\Nodes\Menu\MenuDefinitionLineNode;
 use phpDocumentor\Guides\ParserContext;
+use phpDocumentor\Guides\RestructuredText\Parser\EmbeddedUriParser;
 use phpDocumentor\Guides\RestructuredText\Parser\LinesIterator;
 
+use function array_filter;
 use function array_map;
 
 class ToctreeBuilder
 {
+    use EmbeddedUriParser;
+
     /**
      * @param mixed[] $options
      *
-     * @return ParsedMenuEntryNode[]
+     * @return MenuDefinitionLineNode[]
      */
-    public function buildToctreeFiles(
+    public function buildToctreeEntries(
         ParserContext $parserContext,
         LinesIterator $lines,
         array $options,
     ): array {
-        $toctreeFiles = [];
+        $toctreeEntries = [];
 
-        foreach ($this->parseToctreeFiles($lines) as $file) {
-            $toctreeFiles[] = $file;
+        foreach ($this->parseToctreeEntryLines($lines) as $entry) {
+            $toctreeEntries[] = $entry;
         }
 
-        return $toctreeFiles;
+        return $toctreeEntries;
     }
 
-    /** @return ParsedMenuEntryNode[] */
-    private function parseToctreeFiles(LinesIterator $lines): array
+    /** @return MenuDefinitionLineNode[] */
+    private function parseToctreeEntryLines(LinesIterator $lines): array
     {
-        $linesArray = $lines->toArray();
-        $trimmedLines = array_map('trim', $linesArray);
+        $linesArray =  array_filter(
+            array_map('trim', $lines->toArray()),
+            static fn (string $file): bool => $file !== '',
+        );
 
         $result = [];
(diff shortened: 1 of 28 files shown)


packages/guides-restructured-text/src/RestructuredText/Toc/ToctreeBuilder.php
```diff
@@ -4,39 +4,53 @@
...
             array_map('trim', $lines->toArray()),
             static fn (string $file): bool => $file !== '',
         );
+
+        $result = [];
+        foreach ($linesArray as $line) {
+            $parsed = $this->extractEmbeddedUri($line);
+            $title = $parsed['text'];
+            $reference = $parsed['uri'];
+            $result[] = new ParsedMenuEntryNode($reference, $title);
```
jaapio, line 51: I would like to see this as inline construction, this will save some memory as php is coping all strings. 
```suggestion
            $parsed = $this->extractEmbeddedUri($line);
            $result[] = new ParsedMenuEntryNode($parsed['uri'], $parsed['text']);
```

packages/guides/src/Twig/AssetsExtension.php
```diff
@@ -157,6 +160,10 @@ public function renderMenu(array $context, string $menuType, int $maxMenuCount =
     /** @param array{env: RenderContext} $context */
     public function renderLink(array $context, string $url, string|null $anchor = null): string
     {
+        if (filter_var($url, FILTER_VALIDATE_URL) !== false) {
```
jaapio, line 163: Should this not be part of the urlGenerator logic?

packages/guides/src/Nodes/Menu/ParsedMenuEntryNode.php
```diff
@@ -0,0 +1,38 @@
...
+
+use phpDocumentor\Guides\Nodes\AbstractNode;
+use Stringable;
+
+/**
+ * This node contains the result of parsing the menu entries for a menu.
+ *
+ * @extends AbstractNode<String>
+ */
+final class ParsedMenuEntryNode extends AbstractNode implements Stringable
```
jaapio, line 15: The name of this class is not really describing what kind of node this is, Would `TitledMenuEntryNode` cover the purpose?

jaapio (changes requested):
I would like to see a different name for the `ParsedMenuEntryNode` as it is not really clear to me what the purpose of the node is. Suggestion is to rename it to `TitledMenuEntryNode` or something like `PlainMenuEntryNode` as those would cover the purpose better? Since it's not about a normal document referencing menuEntry?

With the introduction of a new EnternalLinkMenuEntry we can remove some logic that is now spread across different classes to detect the menu item type. And use a test in the template to check how to render a menu item. This would remove the expensive calls to `filter_var($url, FILTER_VALIDATE_URL)` which is way slower than `$item instanceof ExternalLinkMenuEntry`

packages/guides/src/Nodes/Menu/ParsedMenuEntryNode.php
linawolf: The thing is  a Menu entry usually contains the resolved links of the Menu. So I am not really happy with also calling it MenuEntry. But I also have no better Idea...

packages/guides/src/Compiler/NodeTransformers/MenuNodeAddEntryTransformer.php
```diff
@@ -54,11 +60,22 @@ public function leaveNode(Node $node, CompilerContext $compilerContext): Node|nu
         $documentEntriesInTree = [];
         $menuEntries = [];
 
-        foreach ($files as $file) {
+        foreach ($parsedMenuEntryNodes as $parsedMenuEntryNode) {
+            if (filter_var($parsedMenuEntryNode->getReference(), FILTER_VALIDATE_URL) !== false) {
+                $menuEntry = new MenuEntryNode(
```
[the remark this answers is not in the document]
linawolf: Done

packages/guides-restructured-text/src/RestructuredText/Toc/ToctreeBuilder.php
linawolf: yeah, this was a left-over

packages/guides/src/Nodes/Menu/ParsedMenuEntryNode.php
linawolf: How about MenuDefinitionLineNode?

packages/guides/src/Twig/AssetsExtension.php
linawolf: True, moved it there

linawolf:
@jaapio I would like to move the refactoring of how the MenuEntries are created to a follow up. Any other changes needed?

jaapio (approved):
As discussed we will improve in a follow up

PR #793, opened 2023-12-30
Allow Symfony 7

greg0ire:


greg0ire:
@jaapio I believe https://github.com/phpDocumentor/guides/blob/bb3a2914d83ffb27d4db5217dba8ec3a80159fe4/composer.json#L7-L9 prevents properly checking that this works. What should be done?

jaapio:
I think we can remove that as long we make sure we do not commit stuff to the lockfile which is not compatible with php 8.1

greg0ire:
Why do we have a lockfile in the first place? If it's for preventing dev tooling such as phpstan or phpunit from suddenly breaking the build, maybe we should switch to using constraints without `^` or `~` in `composer.json`, and unversioned `composer.lock`?

jaapio:
I'm used to commit the lock file, also for libraries, because the lock file makes sense in terms of making sure everyone working on this project is using the same dependencies as start. You see this in all phpDocumentor projects. 

It does not harm to have it, and it speeds up the initial installation of a project, because composer doesn't have to do the resolving.

greg0ire:
> I think we can remove that as long we make sure we do not commit stuff to the lockfile which is not compatible with php 8.1

This makes contributions a bit more complicated. Instead, I'll try using `--ignore-platform-req=php` where appropriate.


Regarding version of the `composer.lock` I believe the speedup is (now that we use Composer 2) quite negligible:

With `composer.lock`: `composer install  2.02s user 1.54s system 116% cpu 3.065 total`
Without: `composer install  2.52s user 1.59s system 118% cpu 3.465 total`

So you get 500 ms.

By committing  it, you make sure people get a green build when they start contributing, but only locally, since `composer.lock` is disregarded in every job with `highest` or `lowest` dependencies dependencies. It also means there are more jobs than necessary, since you have `locked` jobs. I think you could challenge it, but with `--ignore-platform-req=php`, I think that's not a requirement for this PR :slightly_smiling_face:

greg0ire:
:thinking: hopefully this won't cause any issues for the `lowest` dependencies jobs :crossed_fingers:

greg0ire:
Speaking of speedup: https://github.com/phpDocumentor/guides/actions/runs/7369898764/job/20055701284?pr=793#step:5:85

greg0ire:
Argh, the option I used is not a good solution, because of 8.1-highest. I will try something else.

greg0ire:
Found a solution. Blocked by https://github.com/phpDocumentor/.github/pull/35

jaapio:
Thanks @greg0ire, merged your other pr. Let me know when you need a review.

greg0ire:
@jaapio please review

jaapio:
Thanks!

PR #795, opened 2023-12-30
[FEATURE] add configuration block directive

jaapio:
Add the implementation of the configuration block directive. To make the implemtation more generic a configuration option is introduced. This will allow the users to configure their own labels.

### DIFF
--- a/packages/guides-restructured-text/resources/config/guides-restructured-text.php
+++ b/packages/guides-restructured-text/resources/config/guides-restructured-text.php
@@ -11,6 +11,7 @@ use phpDocumentor\Guides\RestructuredText\Directives\BreadcrumbDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\CautionDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\ClassDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\CodeBlockDirective;
+use phpDocumentor\Guides\RestructuredText\Directives\ConfigurationBlockDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\ConfvalDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\ContainerDirective;
 use phpDocumentor\Guides\RestructuredText\Directives\ContentsDirective;
@@ -115,6 +116,7 @@ use Psr\Log\LoggerInterface;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 use function Symfony\Component\DependencyInjection\Loader\Configurator\inline_service;
+use function Symfony\Component\DependencyInjection\Loader\Configurator\param;
 use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
 use function Symfony\Component\DependencyInjection\Loader\Configurator\tagged_iterator;
 
@@ -175,6 +177,10 @@ return static function (ContainerConfigurator $container): void {
             ),
         ])
         ->set(ConfvalDirective::class)
+        ->set(ConfigurationBlockDirective::class)
+        ->args([
+            '$languageLabels' => param('phpdoc.rst.code_language_labels'),
+        ])
         ->set(ContainerDirective::class)
         ->set(ContentsDirective::class)
         ->arg('$documentNameResolver', service(DocumentNameResolverInterface::class))
--- a/packages/guides-restructured-text/src/RestructuredText/DependencyInjection/ReStructuredTextExtension.php
+++ b/packages/guides-restructured-text/src/RestructuredText/DependencyInjection/ReStructuredTextExtension.php
@@ -8,6 +8,9 @@ use phpDocumentor\Guides\RestructuredText\DependencyInjection\Compiler\TextRoleP
 use phpDocumentor\Guides\RestructuredText\Nodes\ConfvalNode;
 use phpDocumentor\Guides\RestructuredText\Nodes\OptionNode;
 use phpDocumentor\Guides\RestructuredText\Nodes\VersionChangeNode;
+use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -15,19 +18,31 @@ use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 
+use function assert;
 use function dirname;
 use function phpDocumentor\Guides\DependencyInjection\template;
 
-class ReStructuredTextExtension extends Extension implements PrependExtensionInterface, CompilerPassInterface
+final class ReStructuredTextExtension extends Extension implements
+    PrependExtensionInterface,
+    CompilerPassInterface,
+    ConfigurationInterface
 {
     /** @param mixed[] $configs */
     public function load(array $configs, ContainerBuilder $container): void
     {
+        $configuration = $this->getConfiguration($configs, $container);
+        $config = $this->processConfiguration($configuration, $configs);
         $loader = new PhpFileLoader(
             $container,
             new FileLocator(dirname(__DIR__, 3) . '/resources/config'),
         );
 
+        $normalizedLanguageLabels = [];
+        foreach ($config['code_language_labels'] ?? [] as $item) {
+            $normalizedLanguageLabels[$item['language']] = $item['label'];
+        }
+
+        $container->setParameter('phpdoc.rst.code_language_labels', $normalizedLanguageLabels);
         $loader->load('guides-restructured-text.php');
     }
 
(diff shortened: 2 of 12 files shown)


jaapio:
@wouterj  @greg0ire, I copied the work by @wouterj to the project, and adjusted it a bit to allow some customization instead of the fixed list of languages. The added test shows how to configure this. 

Please let me know if this fits your needs.

jaapio:
Maybe we should change the template a bit to work with this css trick: https://css-tricks.com/examples/CSSTabs/radio.php

packages/guides-restructured-text/src/RestructuredText/Directives/ConfigurationBlockDirective.php
```diff
@@ -0,0 +1,81 @@
...
+        private LoggerInterface $logger,
+        Rule $startingRule,
+        array $languageLabels = [],
+    ) {
+        parent::__construct($startingRule);
+
+        $this->slugger = new AsciiSlugger();
+
+        foreach ($languageLabels as $item) {
+            $this->languageLabels[$item['language']] = $item['label'];
```
greg0ire, line 45: :thinking: this doesn't look right. A single language, like `php`, should be able to be associated with several labels, like `attribute` or `annotation`.

packages/guides-restructured-text/resources/config/guides-restructured-text.php
```diff
@@ -173,6 +175,10 @@
             ),
         ])
         ->set(ConfvalDirective::class)
+        ->set(ConfigurationBlockDirective::class)
+        ->args([
+            '$languageLabels' => param('phpdoc.rst.code_language_labels'),
```
greg0ire, line 180: I think I'd use "alias" instead of "label": a single language can be aliased to many different labels. I don't think the opposite needs to be true.

packages/guides-restructured-text/src/RestructuredText/DependencyInjection/ReStructuredTextExtension.php
```diff
@@ -55,4 +65,36 @@ public function process(ContainerBuilder $container): void
     {
         (new TextRolePass())->process($container);
     }
+
+    public function getConfigTreeBuilder(): TreeBuilder
```
wouterj, line 69: Do we want this in the rest extension or the code extension?

The language labels related closely to highlighting aliases, which are configured in the code extension. If we move the tabbed directive to reSt/core, I think we can move configuration block to guides-code (for normal use-cases, you can use the tabbed directive and for code specific docs you can use the guides-code extension).

packages/guides-restructured-text/src/RestructuredText/Directives/ConfigurationBlockDirective.php
wouterj: No, this is correct. We have 2 "maps":

* Alias to highlighting language
* Alias to label

E.g. `.. code-block:: php-attributes` is aliased to `php` in the highlighter, but has the `Attributes` label assigned to it. Maybe we can even combine all of this in one config option (if we move this directive to guides-code), e.g. `<language name="php-attributes" highlight-as="php" label="Attributes"/>`

packages/guides-restructured-text/src/RestructuredText/Directives/ConfigurationBlockDirective.php
```diff
@@ -0,0 +1,81 @@
...
+        }
```
wouterj, line 46: I would move this to the `RestructuredTextExtension::load()` method and make the argument `array<string, string>` (i.e. "array<language, label>`).

wouterj:
> Maybe we should change the template a bit to work with this css trick: https://css-tricks.com/examples/CSSTabs/radio.php

The default theme doesn't come with CSS though as far as I know. Maybe we can use `<details>` as that's the best tab-like view you can create with plain HTML? https://jsfiddle.net/e3zwvp4m/

In the bootstrap extension, we have to override this anyways to use BootstrapCSS's tabs styling. And the same applies to Symfony.

linawolf:
> Maybe we should change the template a bit to work with this css trick: https://css-tricks.com/examples/CSSTabs/radio.php

This "Trick" is not accessible however.

packages/guides-restructured-text/src/RestructuredText/DependencyInjection/ReStructuredTextExtension.php
linawolf: We do not use the code extension but have our own code highlighting. We would however still like to use this directive.

jaapio: Let me check that, because phpDocumentor is also not using the code extension. 

Maybe that extension should be more configurable to make this work. I like the idea to group this functionality as both directives focus on code.

packages/guides-restructured-text/src/RestructuredText/Directives/ConfigurationBlockDirective.php
greg0ire: Oh I see, ok!

packages/guides-restructured-text/src/RestructuredText/DependencyInjection/ReStructuredTextExtension.php
jaapio: I have been thinking about this, the directive is related to the RST format, so it would be strange to put that in the code extension which is not bound to a input format. 

I'm still thinking about a way to make the configuration less of a thing... maybe we should create some kind of mapping between the code extension and the rst extension to make the configuration less of a hassle.

packages/guides/src/Nodes/Configuration/ConfigurationBlockNode.php
```diff
@@ -0,0 +1,24 @@
...
+declare(strict_types=1);
+
+namespace phpDocumentor\Guides\Nodes\Configuration;
+
+use phpDocumentor\Guides\Nodes\AbstractNode;
+
+/** @extends AbstractNode<list<ConfigurationTab>> */
+final class ConfigurationBlockNode extends AbstractNode
+{
+    /** @param list<ConfigurationTab> $tabs */
```
linawolf, line 12: I would give each ConfigurationBlock a unique identifier that can be used to reference the tab parent when rendering the tabs etc

tests/Integration/tests/directives/directive-configuration-block/input/guides.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<guides xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides packages/guides-cli/resources/schema/guides.xsd">
+    <extension class="\phpDocumentor\Guides\RestructuredText\DependencyInjection\ReStructuredTextExtension">
+            <code-language-label language="php" label="CUSTOM" />
```
linawolf, line 6: I would like to see an integration test where ther4e are several labels for the same language

linawolf (approved):
Great work, I only have a few small remarks

packages/guides-restructured-text/src/RestructuredText/Directives/ConfigurationBlockDirective.php
```diff
@@ -0,0 +1,74 @@
...
+            if (!$child instanceof CodeNode) {
+                $this->logger->warning('The ".. configuration-block::" directive only supports code blocks, "' . get_debug_type($child) . '" given.');
+
+                continue;
+            }
+
+            $language = $child->getLanguage();
+            assert($language !== null);
+
+            $label = $this->languageLabels[$language] ?? ucfirst(str_replace('-', ' ', $language));
```
wouterj, line 63: A more generic default might be this:

```suggestion
            $label = $this->languageLabels[$language] ?? $this->slugger->slug($language, ' ')->title()->toString();
```


So that we provide somewhat readable defaults for e.g. `html+php` ("Html php") and `php_attributes` ("Php attributes").

packages/guides/src/Nodes/Configuration/ConfigurationTab.php
```diff
@@ -0,0 +1,22 @@
...
+final class ConfigurationTab
+{
+    public readonly string $hash;
+
+    public function __construct(
+        public readonly string $label,
+        public readonly string $slug,
+        public readonly CodeNode $content,
+    ) {
+        $this->hash = hash('sha1', $content->getValue());
```
wouterj, line 20:
```suggestion
        $this->hash = hash('xxh128', $content->getValue());
```

This is a fast algorithm with less chance of conflict than sha1.

tests/Integration/tests/directives/directive-configuration-block/input/guides.xml
jaapio: I don't think that's part of the feature, the configuration block is intended to have a format per tab. so you cannot have multiple tabs with the same language. 
That's why symfony is using `php+annotations` and `php+attributes`

packages/guides/src/Nodes/Configuration/ConfigurationBlockNode.php
jaapio: how would that look like? an id based on the hashes of all tabs?

linawolf: per document counter. So if you display more then one per page you can still give them unique ids in HTMl?

packages/guides-restructured-text/resources/config/guides-restructured-text.php
linawolf: We only support one tab per language

packages/guides-restructured-text/src/RestructuredText/DependencyInjection/ReStructuredTextExtension.php
linawolf: It is in the rest extension now so that should be fine?

packages/guides/src/Nodes/Configuration/ConfigurationBlockNode.php
linawolf: I think we cna leave that for now

linawolf:
@jaapio shall we merge this?

PR #797, opened 2023-12-31
!!![TASK] Warn only once when a reference cannot be resolved

linawolf:
This is a breaking change to those who have implemented a custom ReferenceResolver.


packages/guides/src/ReferenceResolvers/DelegatingReferenceResolver.php
```diff
@@ -20,21 +21,25 @@ public function __construct(private readonly iterable $resolvers, private readon
     {
     }
 
-    public function resolve(LinkInlineNode $node, RenderContext $renderContext): void
+    public function resolve(LinkInlineNode $node, RenderContext $renderContext, Messages|null $messages = null): void
     {
+        if ($messages === null) {
+            $messages = new Messages();
+        }
```
jaapio, line 28:
```suggestion
        $messages ??= new Messages();
```

jaapio (commented):
Why are the messages nullable? was this an attempt to not make this a breaking change?

packages/guides/src/ReferenceResolvers/DelegatingReferenceResolver.php
linawolf: the DelegatingReferenceResolver is calling all the other resolvers and takes care of the warnings. However it  implements the same Interface. We could however move the message handling to ReferenceResolverPreRender...

linawolf: I moved the Message handling to the PreRenderer. Please have another look

jaapio (approved):

PR #802, opened 2024-01-04
[FEATURE] Load data from composer.json if present

linawolf:
Load data like description, support links, composer name etc from the composer.json if one is found.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,11 @@
         },
         "sort-packages": true
     },
+    "support": {
+        "issues": "https://github.com/phpDocumentor/guides/issues",
+        "source": "https://github.com/phpDocumentor/guides",
+        "docs": "https://docs.phpdoc.org/3.0/guide/guides/index.html"
+    },
     "autoload-dev": {
         "psr-4": {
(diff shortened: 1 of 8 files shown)


jaapio (changes requested):
I would opt for a more generic solution for this feature. If we would pass the projectNode in an event, we could attach any listener to the project node to add extra variables.  

By doing this the run command doesn't have to be aware of the listeners that are part of building step of the project node. And there would be no need to change the Settings as they are just used to transfer the data to the project node? 

If there is a reason to have the composer information in the settings, I would suggest to create a more generic format to store them in the settings. Right now the settings object is really specify changed for the composer addition. And if we have more of these kind of topic it would be grow fast.

linawolf:
How, when and where would you pass the project node?

linawolf:
@jaapio how can we proceed?

jaapio:
> How, when and where would you pass the project node?

I would trigger in event just like you do right now in the run command. Something like `ProjectLoaded` which includes the project node.

linawolf:
@jaapio please have another look if this is what you intended

jaapio (approved):

jaapio:
@linawolf can you please have a look at the merge conflicts?

PR #806, opened 2024-01-06
!![TASK] Improve interlink slug and warning handling

linawolf:
- Add tests
- Improve method names, some have been misleading
- Always Slug anchors
- Do not slug document names (they can be case sensitive etc)

This change is breaking for those who extended inventory handling (that should only be TYPO3).


linawolf:
@jaapio could you review please?

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
```diff
@@ -42,11 +50,7 @@ public function resolve(LinkInlineNode $node, RenderContext $renderContext, Mess
             );
 
             return false;
-        }
-
-        $inventory = $this->inventoryRepository->getInventory($domain);
-        $group = $node instanceof DocReferenceNode ? 'std:doc' : 'std:label';
-        if (!$inventory->hasInventoryGroup($group)) {
+        } catch (InterlinkGroupNotFound) {
```
jaapio, line 53: I would recommend to not use exceptions for cases that are not really exceptions, performance wise exceptions have quit some impact on the execution time in php: https://php.watch/articles/php-exception-performance

It's not an exception that links cannot be found, it's a common flow of the code that a certain link cannot be resolved due misconfiguration.

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
```diff
@@ -29,7 +32,12 @@ public function resolve(LinkInlineNode $node, RenderContext $renderContext, Mess
 
         $domain = $node->getInterlinkDomain();
         $target = $node->getTargetReference();
-        if (!$this->inventoryRepository->hasInventory($domain)) {
+        $group = $node instanceof DocReferenceNode ? 'std:doc' : 'std:label';
+
+        try {
+            $inventory = $this->inventoryRepository->getInventory($domain);
+            $link = $this->inventoryRepository->getLink($domain, $group, $target);
```
jaapio, line 39: Why are this separate methods? Ther inventory contains the links doesn't it?

packages/guides/src/Interlink/Inventory.php
```diff
@@ -16,7 +18,7 @@ final class Inventory
 
     private bool $isLoaded = false;
 
-    public function __construct(private readonly string $baseUrl)
+    public function __construct(private readonly string $baseUrl, private readonly AnchorReducer $anchorReducer)
```
jaapio, line 21: Is it the responsibility of an Inventory to use the AnchorReducer? Isn't that part of the one consuming the links from the inventory group?

jaapio (commented):
I have a few questions and a remark regarding the way you are using exceptions. I must admit that I find it hard to follow what is happening in this part of the code as i'm not very familiar with the interlink code. So please let me know when you think different.

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
linawolf: I learned that it is better to give a reason when something is not found (throw an expected exception) then just return null?

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
linawolf: we need the inventory to get the base url

packages/guides/src/Interlink/Inventory.php
linawolf: Yes because by reducing the anchors before putting them into the inventory and after getting them from the inventory we enforce rules like that anchors are case insensitive and that results with minus and underscore are treated the same.

If this behaviour is not desired - if for example you want case-sensitive anchors - you could still pass a different AnchorReducer to the Inventories

linawolf: I renamed it into anchor reducer

packages/guides/src/ReferenceResolvers/InterlinkReferenceResolver.php
linawolf: I replaced them with the message system, please have another look

linawolf:
@jaapio please have a look if I applied all changes you intended

jaapio (approved):
Thanks this looks very nice

PR #812, opened 2024-01-13
[FEATURE] Make code extension work out-of-the box

linawolf:
* Add default templates
* Use it for our own documentation
* Treat rest and unknown language as "text"
* Use warnings for unknown languages and supply debugging information
* As i also want to use the debugInformation for our third-party highlighter, I introduced a new global into twig which only contains the array. This way third party packages do not have to know about the RenderContext.

### DIFF
--- a/packages/guides-code/resources/template/html/body/code/highlighted-code.html.twig
+++ b/packages/guides-code/resources/template/html/body/code/highlighted-code.html.twig
@@ -1,0 +1,1 @@
+{{ node.value|highlight(node.language) }}
--- a/packages/guides-code/src/Code/DependencyInjection/CodeExtension.php
+++ b/packages/guides-code/src/Code/DependencyInjection/CodeExtension.php
@@ -12,12 +12,13 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\Extension;
+use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 
 use function assert;
 use function dirname;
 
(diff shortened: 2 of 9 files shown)


packages/guides-code/resources/template/html/body/code.html.twig
```diff
@@ -0,0 +1,13 @@
+{% if node.raw %}
+    {# see the RawDirective for where this is coming from; a refactor is desired to move this onto its own template / renderer #}
```
jaapio, line 2: Wondering if this is still an issue. That would clean up this template.

packages/guides-code/resources/template/html/body/code.html.twig
```diff
@@ -0,0 +1,13 @@
...
+    {{ node.value|raw }}
+{%- else -%}
+    {% if node.caption %}
+        <div class="code-block-caption">
+            <span class="caption-text">{{ node.caption }}</span>
+        </div>
+    {%- endif -%}
+    <pre{% if node.classes %} class="{{ node.classesString }}"{% endif %}><code class="language-{{ node.language }}{{ node.startingLineNumber ? ' line-numbers' }}"
+                {%- if node.startingLineNumber %} data-start="{{ node.startingLineNumber }}"{% endif -%}
+                {%- if node.emphasizeLines %} data-emphasize-lines="{{ node.emphasizeLines }}"{% endif -%}>{{ node.value|highlight(node.language) }}</code></pre>
```
jaapio, line 12: The change is just on this line, maybe we should introduce a template block and allow this extension to extend the base template?

jaapio (approved):
Just some remarks regarding the template. We might want to address those in a follow up

packages/guides-code/resources/template/html/body/code.html.twig
linawolf: Not sure, that was already present before I entered the project.

packages/guides-code/resources/template/html/body/code.html.twig
linawolf: that is a good idea, applied it

PR #813, opened 2024-01-14
[TASK] Regenerate integration test baseline

linawolf:
There are only changes of spaces. But they cause noise when you want to run the tool to update integration tests

Additionally fixed the integrationtestcopy tool to ignore skipped directories


linawolf:
Merging as no-brainer

PR #820, opened 2024-01-19
[!!!TASK] Make all classes final

jaapio:
This library has a very large number of extension points via interfaces and abstract classes. All other classes should not be extended by users and shall theirfor be marked as final.

There are a few classes left that should also be marked as final but those require some extra attention.


linawolf (approved):
I think it can save us work if all these classes are final. Thanks for taking care of it!

wouterj:
:+1: 

I see you also made some DTOs final (without interfaces). If these DTOs are part of the public API (either return or parameter type), this can block legitimate extension points (e.g. you can not add new properties to such DTO from a decorator). Not sure if this applies to any of the classes in this PR, but just so you are aware of this.

jaapio:
Thanks @wouterj, I'm aware that this might block extensions, I think we need some extra interfaces, I will review that per case.  However I think some of the helper classes will be marked as internal and should never be exposed to the world as they are part of the internals of this libraries.

PR #826, opened 2024-01-21
Resolve path after checking it leads to an existing file

### LINKED ISSUE #824: Crash when guides-cli is run without a configuration file
I thought I would test 0.3.0 on doctrine, and was greeted with this crash.

```console
$ vendor/bin/guides -vvv --no-progress docs/en
PHP Fatal error:  Uncaught TypeError: is_file(): Argument #1 ($filename) must be of type string, bool given in /home/greg/dev/doctrine-orm/minor/vendor/phpdocumentor/guides-cli/bin/guides:44
Stack trace:
#0 /home/greg/dev/doctrine-orm/minor/vendor/phpdocumentor/guides-cli/bin/guides(44): is_file()
#1 /home/greg/dev/doctrine-orm/minor/vendor/bin/guides(119): include('...')
#2 {main}
  thrown in /home/greg/dev/doctrine-orm/minor/vendor/phpdocumentor/guides-cli/bin/guides on line 44
```

I believe it's caused by https://github.com/phpDocumentor/guides/pull/768 (`realpath` returns `false`, and then `false` is passed to `is_dir`)

greg0ire:
Running the tool without creating a guides.xml is a valid scenario, and should be supported. This means realpath() may return false, which is not a valid input for is_file().

Fixes #824

### DIFF
--- a/packages/guides-cli/bin/guides
+++ b/packages/guides-cli/bin/guides
@@ -40,8 +40,9 @@ if ($input->hasParameterOption('-vvv', true)) {
 
 $containerFactory = new ContainerFactory([new ApplicationExtension()]);
 
-$projectConfig = realpath($vendorDir . '/../guides.xml');
+$projectConfig = $vendorDir . '/../guides.xml';
 if (is_file($projectConfig)) {
+    $projectConfig = realpath($vendorDir . '/../guides.xml');
     if ($verbosity === 3) {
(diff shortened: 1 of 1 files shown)


jaapio (approved):

jaapio:
Thanks for addressing this!

PR #827, opened 2024-01-22
[DOC] configuration.rst

kaystrobach:
* fix broken xsd uri
* change not allowed html-theme to theme

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -23,7 +23,7 @@ building should live in ``guides.xml`` in the current working directory.
     <?xml version="1.0" encoding="UTF-8" ?>
     <guides xmlns="https://www.phpdoc.org/guides"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="https://www.phpdoc.org/guides vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd">
+        xsi:schemaLocation="https://raw.githubusercontent.com/phpDocumentor/guides/main/packages/guides-cli/resources/schema/guides.xsd">
 
         <project title="phpDocumentor Guides"/>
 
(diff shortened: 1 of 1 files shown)


linawolf (changes requested):
Thanks a lot for your contribution! Actually the  xsi:schemaLocation does not have to be an URI and currently points to a the local directory when using the mono-repository. Did you have trouble with it working in your IDE?

About the the attribute html-theme vs theme you are correct, thanks for catching it!

kaystrobach:
Yep, did not work do fetch the xsd, as we do not use the vendor folder in that project. I would prefer to have the xsd centrally stored. This way you can validate it even without the dependencies installed.

linawolf:
I think we had that discussion before and descided for some reason to use relative URLs, can you remember why, @jaapio ?

garvinhicking:
IIRC the main reason was that we figured the XSD would be changed around and we didn't have a URI where the XSD would be published already (and would maybe need release versioning).

For a 1.0 release we should really provide a stable URI somewhere. Maybe if we bundle it within the documentation of `render-guides` and use that "main" version as a stable link?!

kaystrobach:
The uri should contain a version number.

garvinhicking:
What's the best practice here if the XSD is in another repository, where the release versions don't reflect  changes in the xsd itself (i.e. version  2.3.0 of the package uses the same xsd as in version 0.3.0 - so we have new package versions without the xsd changing.
And if the xsd would get a new attribute in version 3.0 of the package that we don't use though, we would need to remember changing all xsd URIs in our docs?

Probably we need to add the xsd to our own domain because of this (=copy it into our own render-guidey repo)?

kaystrobach:
Imho the xsd is a technical description. It should only have mayors as versions. Maybe exceptions for lower versions until 1.0

Maybe you also have „latest“ which can be used in the docs. 🤷‍♂️

linawolf:
Thanks for catching this, @kaystrobach -- and sorry this sat without a response for so long.

Picked up the `html-theme` -> `theme` fix in #1376. The `xsi:schemaLocation` part sparked a real discussion above that never reached a conclusion, so I've captured that separately in #1377 rather than losing it.

PR #833, opened 2024-01-27
Work with relative paths

### LINKED ISSUE #825: Cannot find path "output"
When running `vendor/bin/guides -vvv --no-progress -- docs/en`,  I get 

```
$ vendor/bin/guides -vvv --no-progress -- docs/en

In Run.php line 372:

  [RuntimeException]
  Cannot find path "output".


Exception trace:
  at /home/greg/dev/doctrine-orm/minor/vendor/phpdocumentor/guides-cli/src/Command/Run.php:372
 phpDocumentor\Guides\Cli\Command\Run->getAbsolutePath() at /home/greg/dev/doctrine-orm/minor/vendor/phpdocumentor/guides-cli/src/Command/Run.php:275
 phpDocumentor\Guides\Cli\Command\Run->execute() at /home/greg/dev/doctrine-orm/minor/vendor/symfony/console/Command/Command.php:279
 Symfony\Component\Console\Command\Command->run() at /home/greg/dev/doctrine-orm/minor/vendor/symfony/console/Application.php:1049
 Symfony\Component\Console\Application->doRunCommand() at /home/greg/dev/doctrine-orm/minor/vendor/symfony/console/Application.php:318
 Symfony\Component\Console\Application->doRun() at /home/greg/dev/doctrine-orm/minor/vendor/symfony/console/Application.php:169
 Symfony\Component\Console\Application->run() at /home/greg/dev/doctrine-orm/minor/vendor/phpdocumentor/guides-cli/bin/guides:66
 include() at /home/greg/dev/doctrine-orm/minor/vendor/bin/guides:119
```

This is confusing because I never specified `output` anywhere.

greg0ire:
`getAbsolutePath()` was introduced in #284, but it is unclear to me why. Removing it seems to fix UX inconsistencies without introducing other issues.

It causes non-existent relative paths to be treated differently than non-existent absolute paths: using the former crashes the program, while using the latter results in directories being created at the provided paths.

Closes #825

### DIFF
--- a/packages/guides-cli/src/Command/Run.php
+++ b/packages/guides-cli/src/Command/Run.php
@@ -53,15 +53,12 @@ use Symfony\Component\EventDispatcher\EventDispatcher;
 use function array_pop;
 use function assert;
 use function count;
-use function getcwd;
 use function implode;
 use function is_countable;
 use function is_dir;
 use function microtime;
 use function pathinfo;
-use function realpath;
 use function sprintf;
-use function str_starts_with;
 use function strtoupper;
 
 final class Run extends Command
@@ -261,7 +258,7 @@ final class Run extends Command
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $settings = $this->getSettingsOverriddenWithInput($input);
-        $inputDir = $this->getAbsolutePath($settings->getInput());
+        $inputDir = $settings->getInput();
         if (!is_dir($inputDir)) {
             throw new RuntimeException(sprintf('Input directory "%s" was not found! ' . "\n" .
                 'Run "vendor/bin/guides -h" for information on how to configure this command.', $inputDir));
@@ -281,7 +278,7 @@ final class Run extends Command
         $projectNode = $event->getProjectNode();
         $settings = $event->getSettings();
 
-        $outputDir = $this->getAbsolutePath($settings->getOutput());
+        $outputDir = $settings->getOutput();
         $sourceFileSystem = new Filesystem(new Local($settings->getInput()));
         $sourceFileSystem->addPlugin(new Finder());
         $logPath = $settings->getLogPath();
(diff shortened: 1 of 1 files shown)


linawolf:
I think we needed to introduce this for usage inside of docker. I dont really remember why we introduced it though

packages/guides-cli/src/Command/Run.php
```diff
@@ -261,7 +258,7 @@ private function getSettingsOverriddenWithInput(InputInterface $input): ProjectS
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $settings = $this->getSettingsOverriddenWithInput($input);
-        $inputDir = $this->getAbsolutePath($settings->getInput());
```
jaapio, line 264: I think we should keep this, line, we want to break when input is not an existing directory.

packages/guides-cli/src/Command/Run.php
```diff
@@ -281,7 +278,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $projectNode = $event->getProjectNode();
         $settings = $event->getSettings();
 
-        $outputDir = $this->getAbsolutePath($settings->getOutput());
+        $outputDir = $settings->getOutput();
```
jaapio, line 281: :+1:

jaapio (changes requested):
I think we should just change the output directory handling. Not the input, as the issue was just focusing on output.

packages/guides-cli/src/Command/Run.php
jaapio: Except that the getAbsolutePath did convert it to a real path, relative to the current working directory, I think that's an important difference.

greg0ire: > Except that the getAbsolutePath did convert it to a real path
yes

> relative to the current working directory

Uuuuuh… no? :sweat_smile: I mean it's called `getAbsolutePath`, so, it produced absolute paths. 



Anyway, in the rest of the method `$inputDir` unused from these 3 lines:

```php
        if (!is_dir($inputDir)) {
            throw new RuntimeException(sprintf('Input directory "%s" was not found! ' . "\n" .
                'Run "vendor/bin/guides -h" for information on how to configure this command.', $inputDir));
        }
```

So the only difference it makes is in the error message of that exception. Is it important to have an absolute path in that message?

jaapio (approved):

PR #836, opened 2024-01-28
Revert phpize for versions, never treat them like numbers

### LINKED ISSUE #829: Trailing Zeroes are removed from version number
If I define the version numbers in the guides.xml:

```
    <project
            title="Render guides"
            version="3.0"
            release="3.0.0"
    />
```
Version is treated as a float by the Tree Builder. The Float is later cast into a string and the trailing zero gets lost

wouterj:
Fixes #829 

Don't ask me how/why this works, I just bluntly copied it from @jaapio's work on phpDocumentor's config (https://github.com/phpDocumentor/phpDocumentor/commit/991adcf381140a5900566e2bdb57cef3f8db343c).

### DIFF
--- a/packages/guides/src/DependencyInjection/GuidesExtension.php
+++ b/packages/guides/src/DependencyInjection/GuidesExtension.php
@@ -38,7 +38,10 @@ use function array_values;
 use function assert;
 use function dirname;
 use function is_array;
+use function is_int;
+use function is_string;
 use function pathinfo;
+use function var_export;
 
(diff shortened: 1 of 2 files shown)


stof:
This won't work for `3.1` vs `3.10`

jaapio:
Just tested the case mentioned by @stof, and it does work for 3.10 and 3.1, both are perfectly transformed to their string value. 

The float values are `3.1` and `3.10`  so when doing a var_export they will be converted to their string representation. This is different when doing a cast.

stof:
@jaapio it might make sense to have tests covering this, to ensure that it keeps working in the future.

PR #844, opened 2024-02-11
Bugfix/mdsections

linawolf:


packages/guides/src/Compiler/NodeTransformers/SectionCreationTransformer.php
```diff
@@ -0,0 +1,112 @@
...
+
+                return $lastSection?->getTitle()->getLevel() === 1 ? $lastSection : null;
+            } else {
+                $newSection = new SectionNode($node);
+                if ($lastSection instanceof SectionNode) {
+                    $lastSection->addChildNode($newSection);
+                }
+
+                $this->sectionStack[] = $newSection;
+            }
```
jaapio, line 96: As the previous if block is already returning, this does not need to be an `else`
```suggestion
              $newSection = new SectionNode($node);
              if ($lastSection instanceof SectionNode) {
                  $lastSection->addChildNode($newSection);
              }

              $this->sectionStack[] = $newSection;
            
```

packages/guides/src/Compiler/NodeTransformers/SectionCreationTransformer.php
```diff
@@ -0,0 +1,112 @@
...
+
+        if ($node instanceof TitleNode) {
+            $lastSection = end($this->sectionStack);
+            if ($lastSection instanceof SectionNode && $node !== $lastSection->getTitle() && $node->getLevel() <= $lastSection->getTitle()->getLevel()) {
+                while (end($this->sectionStack) instanceof SectionNode && $node !== end($this->sectionStack)->getTitle() && $node->getLevel() <= end($this->sectionStack)->getTitle()->getLevel()) {
+                    $lastSection = array_pop($this->sectionStack);
+                }
+
+                $newSection = new SectionNode($node);
+                if (end($this->sectionStack) instanceof SectionNode) {
```
jaapio, line 82: Should this not be `$lastSection` rather than `end($this->sectionStack)` ?

packages/guides/src/Compiler/NodeTransformers/SectionCreationTransformer.php
```diff
@@ -0,0 +1,112 @@
...
+        if (count($this->sectionStack) > 0 && $compilerContext->getShadowTree()->isLastChildOfParent()) {
+            $lastSection = end($this->sectionStack);
+            while ($lastSection?->getTitle()->getLevel() > 1) {
+                $lastSection = array_pop($this->sectionStack);
+            }
+
+            return $lastSection;
+        }
+
+        if ($node instanceof TitleNode) {
```
jaapio, line 74: Inverse this logic would make sense to me, as this would make it more readable what happens, only when the current node is a TitleNode we do more complex stuff, This can even be moved 1 if statement up? 
```suggestion
        if ($node instanceof TitleNode === false) {
            return null;
        }
```

jaapio (commented):
This looks very good, I do like how the introduction of a class like this can have impact on how the AST looks like. 

I do have some questions on how we might be able to improve the code a bit, By adding an early exit we might be able to make it a bit more readable, as what I understand the leaveNode is only applicable when the current node is a title node?

packages/guides/src/Compiler/NodeTransformers/SectionCreationTransformer.php
linawolf: nope. $lastSection is the last one that was removed. the end of the stack is the one where further sections need to be attached

PR #849, opened 2024-02-16
[FEATURE] Add line block support

jaapio:
Line blocks are just lines in a div that can be nested using indentation. Every line must start with an | and a block ends when an empty line is detected.

I implemented this using ContainerNodes because it's an rst only type, and it has no other behavior rather than just some custom styling.


linawolf (approved):
Cool, I ve wanted to implement this since a while and never found the time

PR #852, opened 2024-02-19
[FEATURE] Provide Sphinx inventory version 2

linawolf:
for improved interoperability when interlinking from a Sphinx document to a guides document

See also: https://sphobjinv.readthedocs.io/en/stable/syntax.html

### DIFF
--- a/packages/guides/composer.json
+++ b/packages/guides/composer.json
@@ -23,6 +23,7 @@
     "require": {
         "php": "^8.1",
         "ext-json": "*",
+        "ext-zlib": "*",
         "league/flysystem": "^1.1",
         "league/tactician": "^1.1",
         "league/uri": "^6.5 || ^7.0",
--- a/packages/guides/src/Renderer/InterlinkObjectsRenderer.php
+++ b/packages/guides/src/Renderer/InterlinkObjectsRenderer.php
@@ -19,8 +19,11 @@ use phpDocumentor\Guides\RenderContext;
 use phpDocumentor\Guides\Renderer\UrlGenerator\UrlGeneratorInterface;
 
 use function json_encode;
+use function sprintf;
+use function zlib_encode;
 
 use const JSON_PRETTY_PRINT;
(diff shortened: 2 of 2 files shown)


packages/guides/composer.json
```diff
@@ -33,7 +33,8 @@
         "symfony/translation-contracts": "^3.4.1",
         "symfony/http-client": "^6.4.3",
         "twig/twig": "~2.15 || ^3.0",
-        "webmozart/assert": "^1.11"
+        "webmozart/assert": "^1.11",
+      "ext-zlib": "*"
```
garvinhicking, line 37: Indentation here is off, just a minor thing but it catches my eye ;)

packages/guides/src/Renderer/InterlinkObjectsRenderer.php
```diff
@@ -19,8 +19,11 @@
 use phpDocumentor\Guides\Renderer\UrlGenerator\UrlGeneratorInterface;
 
 use function json_encode;
+use function sprintf;
+use function zlib_encode;
```
garvinhicking, line 23: We'll have to see if that zlib module is already configured for the docker-container and the GH Action build?

garvinhicking (commented):
Wow great work getting on it so quickly, thanks!
Two minor things noted.

packages/guides/composer.json
andreaskienast: I'd prefer using composer commands to add dependencies rather than editing the file manually.

packages/guides/src/Renderer/InterlinkObjectsRenderer.php
linawolf: Yes. that would be needed

packages/guides/composer.json
linawolf: In this case I used phpstorm to add the missing extension and didnt double chek

PR #853, opened 2024-02-19
[BUGFIX] Handle "dirty" objects.inv.json data

linawolf:
Handle null values as not set, ignore inventory entries that have non-scalar values or null in the URI.

Until now those were causing type errors

### DIFF
--- a/packages/guides/src/ReferenceResolvers/Interlink/DefaultInventoryLoader.php
+++ b/packages/guides/src/ReferenceResolvers/Interlink/DefaultInventoryLoader.php
@@ -20,6 +20,9 @@ use Symfony\Component\HttpClient\Exception\ClientException;
 
 use function count;
 use function is_array;
+use function is_scalar;
+use function is_string;
+use function sprintf;
 use function strval;
 
 final class DefaultInventoryLoader implements InventoryLoader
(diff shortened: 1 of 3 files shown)


brotkrueml (commented):
Sorry, currently I don't have enough information for a review: Which content holds `$linkArray`?

linawolf:
$linkData comes from the imported JSON data, see the fixtures in the Unit Tests https://github.com/phpDocumentor/guides/pull/853/files#diff-615ca620b21221a05c303d924a986d8334be1c95db29e038e95c7a23d60f21b1

packages/guides/src/ReferenceResolvers/Interlink/DefaultInventoryLoader.php
```diff
@@ -46,11 +48,22 @@ public function loadInventoryFromJson(Inventory $inventory, array $json): void
...
+                        continue;
+                    }
+
+                    if (!is_string($linkArray[2])) {
+                        $this->logger->warning('Invalid Inventory entry found. ');
+                        continue;
+                    }
+
+                    if (!is_scalar($linkArray[0] ?? '') || !is_scalar($linkArray[1] ?? '') || !is_scalar($linkArray[3] ?? '')) {
+                        $this->logger->warning('Invalid Inventory entry found. ');
```
garvinhicking, line 61: Maybe we should emit unique errors for each string here, so that later we easier know what exact error is triggered?

linawolf: Done

PR #855, opened 2024-02-19
[DOCS] Document the Renderer

linawolf:


docs/components/renderer.rst
```diff
@@ -6,5 +6,51 @@
...
+:php:class:`\phpDocumentor\Guides\Renderer\BaseTypeRenderer` and is
+empty otherwise.
+
+..  literalinclude:: _renderer/_PlaintextRenderer.php
+    :language: php
+    :caption: src/Renderer/PlaintextRenderer.php
+
+Register the new renderer in the container:
+
+..  literalinclude::_renderer/_myextension.php
```
garvinhicking, line 37: I believe here's a space missing after "::"

docs/components/renderer.rst
```diff
@@ -6,5 +6,51 @@
...
+
+All renderers must be registered in the ContainerConfigurator of the extension
+with the tag :php:`'phpdoc.renderer.typerenderer'` and additional format information.
+
+Example: a plaintext renderer
+=============================
+
+Create a class called :php:`PlaintextRenderer` which extends
+:php:class:`\phpDocumentor\Guides\Renderer\BaseTypeRenderer` and is
+empty otherwise.
```
garvinhicking, line 29: I suppose the reason for an empty renderer is that the BaseTypeRenderer already renders als plaintext? Maybe some note about this like "and is empty otherwise (because the BaseTypeRenderer will provide a raw unaltered rendering of the AST)"?

docs/components/renderer.rst
```diff
@@ -6,5 +6,51 @@
...
+    <?xml version="1.0" encoding="UTF-8" ?>
+    <guides xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides packages/guides-cli/resources/schema/guides.xsd"
+    >
+        <extension class="MyVendor\MyExtension"/>
+        <output-format>txt</output-format>
+    </guides>
+
+You can now define custom templates or even custom NodeRenderer for the new output
```
jaapio, line 55: It would be nice to explain how this is done, or link to a section where this is already covered, not sure we have one.

docs/components/renderer.rst
```diff
@@ -6,5 +6,51 @@
 Renderer
 ========
 
-..  warning::
-    This needs to be documented
+A Renderer transforms the :abbreviation:`AST (abstract syntax tree)` into a
```
brotkrueml, line 9:
```suggestion
A Renderer transforms the :abbr:`AST (abstract syntax tree)` into a
```

At least in Sphinx it is abbr.

But I would use just `Abstract Syntax Tree (AST)` when mentioning first, so no need to hover it.

docs/components/renderer.rst
brotkrueml, line 29: Actually, I don't understand the "and is empty otherwise". Why otherwise? I wouldn't write "empty", maybe better "which just extends BaseTypeRenderer" and omit the "empty otherwise".

docs/components/renderer.rst
linawolf: The two textroles are synonyms

docs/components/renderer.rst
linawolf: good point!

docs/components/renderer.rst
linawolf: They have yet to be written :-(

PR #856, opened 2024-02-19
Use branch alias to set version

jaapio:


linawolf (approved):

linawolf:
I think it is a good idea

PR #867, opened 2024-02-24
[TASK] Ignore domains with unknows directives and textroles

linawolf:


jaapio:
Does this also enable known directives and roles in a certain domain? I guess when I read this code it doesn't, so it ignores only the unknown?

linawolf:
For now it only disables unknown directives with domain. We do have no central domain handling in place right now. But introducing such would be beyond the scope of this PR. I think we need to give it more thought.

This PR enables us to check for warnings in the docs again and therefore prevent other kind of warnings

jaapio (approved):

PR #871, opened 2024-02-26
[FEATURE] Introduce NavigationTitle

linawolf:
In TYPO3 we often have the problem, that longer titles are required for search engines, our internal search and for ppl  coming from by direkt link out of context. However long titles clutter up the menues and cause a lot of repetition there.

So we would like to be able to give Documents a short title for the Menu and a longer one to display in place.

This is not a sphinx feature, it is however hard to impossible to do in a theme


jaapio:
In some spots this feels like a dirty hack, I see that you are using a field to set the alternative navigation title. That looks good. what feels wrong is that we need an extra option in the DocumentEntry. I would expect the same approach as for the breadcrumbs where you chose on rendering the different title. I think we should do the same for menu's. 

The document node should not be aware of the existing extra title. Maybe we should have an extra map where you can do the title lookup to make it easier. we could build that map during the compilingpass and use that during the rendering phase. 

I hope this makes sense? Let's discuss the details if you want to know more.

linawolf:
@jaapio how would you feel about this solution, where I keep the data in a map and make it extendable by event?

PR #900, opened 2024-03-04
[BUGFIX] Display external pages on all levels of the main menu

linawolf:
Sphinx handels it also that way

### DIFF
--- a/packages/guides/src/Compiler/Passes/GlobalMenuPass.php
+++ b/packages/guides/src/Compiler/Passes/GlobalMenuPass.php
@@ -17,10 +17,13 @@ use phpDocumentor\Guides\Compiler\CompilerContext;
 use phpDocumentor\Guides\Compiler\CompilerPass;
 use phpDocumentor\Guides\Nodes\DocumentNode;
 use phpDocumentor\Guides\Nodes\DocumentTree\DocumentEntryNode;
+use phpDocumentor\Guides\Nodes\DocumentTree\ExternalEntryNode;
+use phpDocumentor\Guides\Nodes\Menu\ExternalMenuEntryNode;
 use phpDocumentor\Guides\Nodes\Menu\InternalMenuEntryNode;
 use phpDocumentor\Guides\Nodes\Menu\MenuEntryNode;
 use phpDocumentor\Guides\Nodes\Menu\NavMenuNode;
 use phpDocumentor\Guides\Nodes\Menu\TocNode;
+use phpDocumentor\Guides\Nodes\TitleNode;
 use phpDocumentor\Guides\Settings\SettingsManager;
 use Throwable;
 
@@ -118,7 +121,7 @@ final class GlobalMenuPass implements CompilerPass
     private function addSubEntries(
         CompilerContext $compilerContext,
         MenuEntryNode $sectionMenuEntry,
-        DocumentEntryNode $documentEntry,
+        DocumentEntryNode|ExternalEntryNode $entryNode,
         int $currentLevel,
         int $maxDepth,
     ): void {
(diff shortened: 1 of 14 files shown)


jaapio:
I will have to review this more in depth, but it looks like there are a number of breaking changes in this pr. 

Will come back to this later

linawolf:
All places that use now DocumentEntryNode|ExternalEntryNode before used DocumentEntryNode so it shouldn't really be breaking.

Also as the Transformers are final they should not have been extended so we do not need to worry there.

I introduced a new level of inheritence in Nodes, however now the DocumentEntryNode extends the EntryNode which extends the AbstractNode, before it extended the AbstractNode directly. This should also be not breaking.

There is a slight possibility that someone expects all children of  DocumentEntryNode to be  DocumentEntryNode. And now this only returns an EntryNode. We could instead of altering getChildren introduce a new method and let getChildren filter only for DocumentEntryNodes?

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
```diff
@@ -118,7 +121,7 @@ private function getMenuEntryWithChildren(CompilerContext $compilerContext, Menu
     private function addSubEntries(
         CompilerContext $compilerContext,
         MenuEntryNode $sectionMenuEntry,
-        DocumentEntryNode $documentEntry,
+        DocumentEntryNode|ExternalEntryNode $entryNode,
```
jaapio, line 124: We can change this to `EntryNode` as argument, as both classes extend this node type. Stretching the accepted types isn't a breaking change in a concreet implementation. like this class, especially because it's final.

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
```diff
@@ -130,17 +133,31 @@ private function addSubEntries(
...
+                    $subEntryNode->getFile(),
+                    $subEntryNode->getTitle(),
+                    [],
+                    false,
+                    $currentLevel,
+                    '',
+                );
+                $sectionMenuEntry->addMenuEntry($subMenuEntry);
+                $this->addSubEntries($compilerContext, $subMenuEntry, $subEntryNode, $currentLevel + 1, $maxDepth);
+            } elseif ($subEntryNode instanceof ExternalEntryNode) {
```
jaapio, line 152: Please separate these blocks as they are not connected, maybe we should extract both strategies to a different method in this class to make it easier to read?

```suggestion
                $this->addSubEntries($compilerContext, $subMenuEntry, $subEntryNode, $currentLevel + 1, $maxDepth);
                continue;
            } 
            
            if ($subEntryNode instanceof ExternalEntryNode) {
```

packages/guides/src/Nodes/DocumentTree/DocumentEntryNode.php
```diff
@@ -38,27 +36,17 @@ public function getTitle(): TitleNode
         return $this->titleNode;
     }
 
-    public function addChild(DocumentEntryNode $child): void
+    public function addChild(DocumentEntryNode|ExternalEntryNode $child): void
```
jaapio, line 39: Please use `EntryNode` here. This should not break existing implementations.

packages/guides/src/Nodes/DocumentTree/DocumentEntryNode.php
```diff
@@ -38,27 +36,17 @@ public function getTitle(): TitleNode
...
     {
         $this->entries[] = $child;
     }
 
-    /** @return DocumentEntryNode[] */
+    /** @return array<DocumentEntryNode|ExternalEntryNode> */
```
jaapio, line 44: Here is an issue, we are stretching the return type. What could help is to add an argument to the getChildren method to specify the filtered child types. Like this:

```
getChildren(string|null $childType = DocumentEntryNode::class) {
```

if we specify null, we can still get all children which we can use internally. Otherwise it will fall back to the original behavior and just provide you an array of `DocumentEntryNode`s. We could directly deprecate this filtering and mention that we will remove this in v2. 


Another approach would be to make sure `ExternalEntryNode has exactly the same interface as the `DocumentEntryNode`   But that would require you to add the methods `getSections`, `getFile` and `isRoot` to `EnternalEntryNode`. Which will be odd, as it does not work like that :-)

jaapio (changes requested):
To be fair, we just released 1.0.0. So I do not expect anyone to use this functionality in to much depth, and the breaking change is limited. But I would like to get some awareness around the things that have been changed. 

Stretching the arguments of methods is ok, people using the methods directly will not be impacted as we still accept the same types as before. The changes in the compiler passes should also not really impact people. 

Changing return types is a no go, as I explained in the review comments. I would like to be constructive in this and not to strict. But I also want to ensure that we are all strict on these type of changes.

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
linawolf: Using EntryNode I had trouble with PHP-stan that I could not resolve...

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
linawolf: What exactly do you mean with "separate the blocks"?

packages/guides/src/Nodes/DocumentTree/DocumentEntryNode.php
linawolf: Cannot happyfy phpstan :-(

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
linawolf: also we are explicitly handling these two types while there could be others...?

packages/guides/src/Nodes/DocumentTree/DocumentEntryNode.php
linawolf: On investigation it might also make sense to keep getChildren as is this way we dont have to filter in places where we only need the child documents. 

And external menu entries are not really child documents anyway

linawolf: done

linawolf:
@jaapio I have some questions, see above

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
jaapio: the `if` and `elseif` are called code blocks, basically everything within `{}` is a code block. A refactor strategy is to always extract blocks, some people state that you should only have 1 level of blocks in a method. I do not want to go this far. but as this is like a strategy it would make sense to make it easy to see what is handled. 

Extracting the strategies to different methods helps to read this method, as it would simply read like

```
if (a) -> doSomething
if (b) -> doSomethingElse
```

packages/guides/src/Compiler/Passes/GlobalMenuPass.php
jaapio: What was the error?

jaapio (approved):

PR #907, opened 2024-03-06
[FEATURE] Support list-table directive

### LINKED ISSUE #890: Support "list-table" directive
### Feature request

The `list-table` directive is currently not supported. It would be nice to have this also in guides:
https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table

linawolf:
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html

resolves #890

@jaapio can we still backport features?

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/ListRule.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Parser/Productions/ListRule.php
@@ -173,8 +173,6 @@ final class ListRule implements Rule
         }
 
         if (!isset($nodes[0])) {
-            $this->logger->warning('List item without content', $blockContext->getLoggerInformation());
-
             return $listItem;
         }
 
--- a/packages/guides/src/Compiler/NodeTransformers/ListNodeTransformer.php
+++ b/packages/guides/src/Compiler/NodeTransformers/ListNodeTransformer.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of phpDocumentor.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @link https://phpdoc.org
+ */
+
+namespace phpDocumentor\Guides\Compiler\NodeTransformers;
+
+use phpDocumentor\Guides\Compiler\CompilerContext;
+use phpDocumentor\Guides\Compiler\NodeTransformer;
+use phpDocumentor\Guides\Nodes\ListItemNode;
+use phpDocumentor\Guides\Nodes\ListNode;
+use phpDocumentor\Guides\Nodes\Node;
+use Psr\Log\LoggerInterface;
+
+use function assert;
+
+/** @implements NodeTransformer<ListNode> */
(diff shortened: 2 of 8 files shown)


jaapio:
Yes for sure as they are not breaking changes we can

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/ListRule.php
```diff
@@ -59,7 +58,6 @@ final class ListRule implements Rule
 
     public function __construct(
         private readonly RuleContainer $productions,
-        private readonly LoggerInterface $logger,
```
jaapio, line 62: I thinks we should not remove this right now, it's not part of the change, and I see removing an argument as a breaking change. Also because a change in the tests is required. If this leads to an unused property, we could transform it to a normal parameter, and make it optional. 
Add a deprecation and remove it in v2.

packages/guides-restructured-text/tests/unit/Parser/Productions/ListRuleTest.php
```diff
@@ -26,7 +25,7 @@ final class ListRuleTest extends RuleTestCase
     protected function setUp(): void
     {
         $ruleContainer = $this->givenCollectAllRuleContainer();
-        $this->rule = new ListRule($ruleContainer, new NullLogger());
+        $this->rule = new ListRule($ruleContainer);
```
jaapio, line 28: please leave the tests untouched, whenever you need to change a unittest that might indicate a breaking change :-)

packages/guides/src/Compiler/NodeTransformers/ListNodeTransformer.php
```diff
@@ -0,0 +1,62 @@
...
+    public function leaveNode(Node $node, CompilerContext $compilerContext): Node|null
+    {
+        assert($node instanceof ListNode);
+        foreach ($node->getChildren() as $listItemNode) {
+            assert($listItemNode instanceof ListItemNode);
+            if (!empty($listItemNode->getChildren())) {
+                continue;
+            }
+
+                        $this->logger->warning('List item without content', $compilerContext->getLoggerInformation());
```
jaapio, line 47: Indending seems to be of here.

jaapio (commented):
Just some minor changes, rest looks good. Thanks for adding this!

packages/guides-restructured-text/src/RestructuredText/Parser/Productions/ListRule.php
linawolf: I just added it to the phpstan baseline. we might need it again , there is no harm in leaving it in

packages/guides-restructured-text/tests/unit/Parser/Productions/ListRuleTest.php
linawolf: done

packages/guides/src/Compiler/NodeTransformers/ListNodeTransformer.php
linawolf: done

jaapio (approved):

PR #910, opened 2024-03-08
[TASK] Improve menu rendering speed

jaapio:
By introducing a cache array for menus we do improve the speed when menus are multiple times rendered.


linawolf (approved):
Wow thanks, this will be a great imptovement

Issue #911, opened 2024-03-08
Render `target` property for image directive


brotkrueml:
### Feature request

The `image` directive can have a `target` to link the image:

```rst
.. image:: /Images/SlidesCover.jpg
   :target: https://example.org/
```

For example, as used here: https://docs.typo3.org/p/ttn/tea/3.0/en-us/Introduction.html#presentation-at-the-typo3-online-days-2021

Reference: https://docutils.sourceforge.io/docs/ref/rst/directives.html#image

Maybe also other properties for `image` are missing, haven't checked.

PR #914, opened 2024-03-09
[FEATURE] Support directive image with target

linawolf:
resolves https://github.com/phpDocumentor/guides/issues/911


jaapio:
Would it not make more sense rather than having a image with a link to wrap the image in a hyperlink when the target is defined? This would allow you to reuse such hyperlinknode, and apply it to other situations. And it would also allow compiler passes to remove the target more easily.

linawolf:
The InlineLinkNodes only accept text as content

PR #930, opened 2024-03-15
[BUGFIX] Fix non-working short-options for "config" and "working-dir"

garvinhicking:
This adresses:

* OutputInterface namespace import is unused
* realpath() may fail, needs gating to satisfy phpstan even though it's unlikely that is_file() returns true, but realpath() doesn't.
* Currently only `--config` and not `-c` is parsed, same with `--working-dir` (`-w`)
* Note: A phpstan error remains due to `getcwd()` maybe returning false and I didn't want to obfuscate this statement with if-condititions; if you have suggestions to improve the type juggling here, tell me.

### DIFF
--- a/packages/guides-cli/bin/guides
+++ b/packages/guides-cli/bin/guides
@@ -9,7 +9,6 @@ use Symfony\Component\Console\Input\ArgvInput;
 use phpDocumentor\Guides\Cli\Application;
 use phpDocumentor\Guides\Cli\DependencyInjection\ApplicationExtension;
 use phpDocumentor\Guides\Cli\DependencyInjection\ContainerFactory;
-use Symfony\Component\Console\Output\ConsoleOutput;
 
 $vendorDir = dirname(__DIR__) . '/../../vendor';
 $autoloadDirectory = $vendorDir . '/autoload.php';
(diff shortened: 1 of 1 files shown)


linawolf (approved):
Thanks for these changes, Looks good to me

packages/guides-cli/bin/guides
```diff
@@ -47,11 +46,13 @@ if (is_file($projectConfig)) {
         echo 'Loading guides.xml from ' . $projectConfig . PHP_EOL;
     }
     // vendor folder was placed directly into the project directory
-    $containerFactory->addConfigFile($projectConfig);
+    if ($projectConfig) {
```
jaapio, line 49: Please make this a more explicit check. I think this check should be !== False?

jaapio (changes requested):

packages/guides-cli/bin/guides
garvinhicking: I did a lax check on that to also evalute an empty string to false.

But I can change that to:

```
if ($projectConfig !== false && $projectConfig !== '')
```

(but that has more cognitive load to me...)

jaapio: That's why I choose to use early returns as much as possible. which is a bit hard in this context as we do not have real functions in this file.  We could abuse `empty($projectConfig) === false` here. or a small namespaced function:

```php
namespace phpDocumentor/guides {
  function emptyOption(string|bool $path) {
     if ($path === false) {
        return true;
    }
  }
} 
```

garvinhicking: I'd be happy to implement it anyway you prefer, I don't have strong feelings about either approach. Personally I'd then rather use the `if ($projectConfig !== false && $projectConfig !== '')` approach, because I find the namespace variant and the empty(...) variant even harder to quickly read 😇

... sidequest ahead...

I did try to optimize this actual code block in the CLI binary because the current way of loading and overloading guides.xml created a problem for us in render-guides. And it's really not a beautiful piece of loading code ;) 

I wanted to actually adjust that "guides.xml" could be fetched from the default render location (which is "docs" in `guides` and "Documentation" in `render-guides`). 

But I couldn't get that to work because of the tough mixture of Decorator and actual Console command and loading order (see https://github.com/TYPO3-Documentation/render-guides/issues/421#issuecomment-1999565902). I abandoned that for the time being, since we got bigger fish to fry at this point than putting hours into this optimization ;)

jaapio:
Thanks a lot @garvinhicking!

jaapio (approved):

PR #931, opened 2024-03-16
[BUGFIX] Prevent Twig from swallowing whitespace between inline tags

### LINKED ISSUE #922: [Bug]: Space missing after bold followed by a reference
### Summary

The following rst:

```rst
**all** :ref:`data <tce-data>`
```
produces the following code:

```html
<strong>all</strong><a href="../Database/Index.html#tce-data">data</a>
```

Obviously, a space is missing after `</strong>`.

The same happens for italic:

```rst
*all* :ref:`data <tce-data>`
```

See summary

### Expected output

A space between the closing tag and the link

linawolf:
resolves #922

I can only solve this by outputting &32; which equals a space. @wouterj is there any solution to force Twig to preserve a real space char?

### DIFF
--- a/packages/guides/resources/template/html/inline/plain-text.html.twig
+++ b/packages/guides/resources/template/html/inline/plain-text.html.twig
@@ -1,1 +1,5 @@
-{{- node.value -}}
+{%- if node.value == ' ' -%}
+    {{ '&#32;'|raw }}
+{%- else -%}
+    {{ node.value }}
+{%- endif -%}
--- a/tests/Functional/tests/definition-list/definition-list.html
+++ b/tests/Functional/tests/definition-list/definition-list.html
@@ -31,7 +31,7 @@
             <span class="classifier"><code>classifier</code></span>
         </dt>
         <dd>Definition 1</dd>
-        <dt><strong>Complex:</strong> <code>int</code></dt>
+        <dt><strong>Complex:</strong>&#32;<code>int</code></dt>
         <dd>Colon must be surrounded by spaces to be a term identifier</dd>
         <dt>multi-line definition term</dt>
         <dd>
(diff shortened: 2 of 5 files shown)


wouterj:
> I can only solve this by outputting &32; which equals a space. @wouterj is there any solution to force Twig to preserve a real space char?

It does not appear to be Twig who swallows the whitespace. If I try out this minimal example, the output is `·` (i.e. Twig renders it as a single space):

```php
<?php

require_once 'vendor/autoload.php';

$loader = new \Twig\Loader\ArrayLoader([
    'test.twig' => '{{- value -}}',
]);
$twig = new \Twig\Environment($loader);

/** Small function to make whitespace visible */
function show_whitespace(string $text) {
    return preg_replace(['/\v/', '/\h/'], ["¶\n", '·'], $text);
}

echo show_whitespace(
  $twig->render('test.twig', ['value' => ' '])
);
```

I've always been a bit confused by how whitespace is handled in this project, the need for `{% apply spaceless %}` in some templates is also strange to me. I think we must identify which part of the NodeRenderer process is responsible for swallowing the whitespace.

linawolf:
I can render any sign but a space. I have no idea how twig works. Chatgbt told me it tries to get rid of superfluous spaces...

linawolf:
applyspaceless is needed in inline templates, because our id adds some such as an empty line at the end of each document. These would then get output, however **bold**: then becomes **bold** :

wouterj:
I'll find some time this weekend to see if I can find the troubling spot using the test case you added in this PR.

wouterj:
I've tracked down this issue to be a result of using the `spaceless` filter in `paragraph.html.twig`. I've refactored whitespace control to avoid using `spaceless` anywhere and taken your test from here in #944. We can close this as being replaced by the new PR.

Issue #924, opened 2024-03-13
Make confvals linkable again


linawolf:
### Feature request

Make confvals linkable again

brotkrueml:
That also means, that a confval must be unique (to determine the link destination) and duplicate confval definitions must output a warning (like in Sphinx). This is currently not the case.

linawolf:
I am working on some idea

PR #932, opened 2024-03-16
[BUGFIX] Make links to special objects prefixable

linawolf:
If we have both an anchor like '.. _demo:' and
a confval of the same name we would get duplicate id's and the links to those two elements could not be distinguished. Sphinx solves this by prefixing all confval links with "confval-".

Therefore I introduces the possibility of prefixing links to destinguished linkable objects.

related https://github.com/phpDocumentor/guides/issues/924

releases: main, 1.0


packages/guides/src/ReferenceResolvers/AnchorReferenceResolver.php
```diff
@@ -46,7 +46,7 @@ public function resolve(LinkInlineNode $node, RenderContext $renderContext, Mess
             return false;
         }
 
-        $node->setUrl($this->urlGenerator->generateCanonicalOutputUrl($renderContext, $target->getDocumentPath(), $target->getAnchor()));
+        $node->setUrl($this->urlGenerator->generateCanonicalOutputUrl($renderContext, $target->getDocumentPath(), $target->getPrefix() . $target->getAnchor()));
```
jaapio, line 49: This seems a bug to me, as you already prefix the anchor in the `getAnchor`?

jaapio (commented):
Wauw... what an ugly hack to make this work. But as it is compatible with sphinx let's do it this way. 

I do wonder if I saw something that's not working as expected but it's a bit hard to see just based on the code.

packages/guides/src/ReferenceResolvers/AnchorReferenceResolver.php
linawolf: The internal target saves the prefix separately. Within rst we do not use the prefixes for references, only in the rendering of links and ids they are prefixed

Issue #942, opened 2024-03-17
Make an integration test that raw does not work


linawolf:
### Feature request

Make an integration test that raw does not work

PR #943, opened 2024-03-17
!!![TASK] Breaking: Disallow raw directive

linawolf:
Using the raw directive is a potencial securtiy risk.

resolves:#942
releases: main, 1.0

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Directives/RawDirective.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Directives/RawDirective.php
@@ -13,34 +13,27 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Guides\RestructuredText\Directives;
 
-use phpDocumentor\Guides\Nodes\Node;
-use phpDocumentor\Guides\Nodes\RawNode;
 use phpDocumentor\Guides\RestructuredText\Parser\BlockContext;
 use phpDocumentor\Guides\RestructuredText\Parser\Directive;
-
-use function implode;
+use Psr\Log\LoggerInterface;
 
 /**
- * Renders a raw block, example:
- *
- * .. raw::
- *
- *      <u>Underlined!</u>
- *
- * @link https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw-data-pass-through
+ * This directive is deactivated for security reasons. If you need it in your project, you must implement it yourself.
  */
-final class RawDirective extends BaseDirective
+final class RawDirective extends ActionDirective
 {
+    public function __construct(
+        private readonly LoggerInterface $logger,
+    ) {
+    }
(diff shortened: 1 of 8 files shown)


packages/guides-restructured-text/src/RestructuredText/Directives/RawDirective.php
```diff
@@ -13,34 +13,27 @@
...
-    /** {@inheritDoc} */
-    public function process(
-        BlockContext $blockContext,
-        Directive $directive,
-    ): Node|null {
-        return new RawNode(implode("\n", $blockContext->getDocumentIterator()->toArray()));
+    public function processAction(BlockContext $blockContext, Directive $directive): void
+    {
+        $this->logger->error('The raw directive is not supported for security reasons. ', $blockContext->getLoggerInformation());
     }
```
wouterj, line 38: I'm not 100% convinced here. I see a security issue when using reStructured Text for "user" input. For projects like the Symfony docs, where the only parsed docs are text maintained by the doc team, there is no security issue. I guess this applies to most usages of Guides.

Implementing the raw node also seems to be non-trivial. One needs to learn how to create a custom Guides extension and how to create a custom directive.

If we want this, I think we should do 2 things:
- Be more explicit about the security issue the raw node causes, so users can make informed decisions.
- Add a config option to disable the raw node, like [docutils](https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw-data-pass-through) also has.

We then have to think about the default, docutils defaults to `true` which we might want to follow for compatibility reasons. Or we can decide to default to `false` if we think the security issues are common enough to protect users against.

jaapio: I would not remove this directive from the code base but allow a config option to enable it. Or log the error when disabled. By making it an option people could still overwrite the behavior using configuration. but this could be enforced by a listener in an extension to ensure people cannot overwrite the configuration. 

The security risk here is on the typo3 side, as the documentation is published on their website, but maintained by approved third-party developers.

wouterj: > The security risk here is on the typo3 side, as the documentation is published on their website, but maintained by approved third-party developers.

We have a similar situation for some community bundles where we render their docs in https://symfony.com/bundles . So this config option would be useful for us, although we need the raw directive for the Symfony docs.

linawolf: I think when you have to use the raw directive that means, that there is something missing in your pipline. It is also very hard to change such HTML output automatically lateron. So I would rather suggest to write specialized directives for your use cases...

linawolf:
I will make another PR

PR #944, opened 2024-03-17
Fix whitespace control

### LINKED ISSUE #922: [Bug]: Space missing after bold followed by a reference

wouterj:
Fixes #922 

Some templates used `{% apply spaceless %}`. This filter does a [very naive preg_replace](https://github.com/twigphp/Twig/blob/b46e93c7257fb01b7c77768210997b1e00643b91/src/Extension/CoreExtension.php#L1132-L1135) to remove any space between HTML tags. This is a big issue, as it also removes whitespace between inline HTML tags (as shown by the testcase added in 5ede52dcde62b68db57c063429b80a1590ce719c).

We must no longer use spaceless in templates, as this creates issues like this. Instead, we must rely on Twig's whitespace control characters: `{{-`/`{%-`/`{#-` removing all whitespace before/after the Twig tag and  `{{~`/`{%~`/`{#~` removing all horizontal whitespace before/after the Twig tag.

I also introduced a Twig file loader that removes the last newline feed from files (8790e8e129b14fd3cdddbc16879df37a5774345a). In unix systems, all files must end with a new line feed. But we don't want all templates to end with a new line feed in the output. That seems to be the main reason spaceless was used in this project. By removing them during loading, we no longer need spaceless. If you want to enforce a new line in the resulting HTML, you can end the file with a Twig comment (as the new line between the last HTML and the Twig comment is not removed).

Furthermore, I improved whitespace control a bit to create nicer output HTML and had to update a LOT of tests (c1452ed7239d13536082bc56eb91f0ac1e7d3c1c). Merging this PR might also result in tests breaking in TYPO3 / Symfony Docs test suites, but this would have been the case also when only removing the use of spaceless.
(quick note: I updated most tests manually, except from the `tests-full` tests as these are very tedious. We might want to double check that all whitespace changes in this file are indeed not resulting in a display change)

### DIFF
--- a/packages/guides-theme-bootstrap/resources/template/body/directive/tabs/tabs-body.html.twig
+++ b/packages/guides-theme-bootstrap/resources/template/body/directive/tabs/tabs-body.html.twig
@@ -1,3 +1,4 @@
 <div class="tab-pane fade {%- if tab.active %} show active{% endif -%}" id="{{ tab.key }}-{{ tabsId }}" role="tabpanel" aria-labelledby="{{ tab.key }}-tab-{{ tabsId }}">
     {{ renderNode(tab.value) }}
 </div>
+{# force a new line at the end of the file #}
--- a/packages/guides-theme-bootstrap/resources/template/body/directive/tabs/tabs-button.html.twig
+++ b/packages/guides-theme-bootstrap/resources/template/body/directive/tabs/tabs-button.html.twig
@@ -6,3 +6,4 @@
         {{ renderNode(tab.content) }}
     </button>
 </li>
+{# force a new line at the end of the file #}
--- a/packages/guides-theme-bootstrap/resources/template/body/menu/mainmenu/menu-level.html.twig
+++ b/packages/guides-theme-bootstrap/resources/template/body/menu/mainmenu/menu-level.html.twig
@@ -1,17 +1,19 @@
 
 <ul class="level-{{ menuEntry.level }}">
-        {% for entry in menuEntry.children -%}
-                <li><a href="{{ renderLink(entry.url) }}"
-                       class="nav-link {%- if menu.currentPath == entry.url %} current {%- endif -%}{% if entry.url in menu.rootlinePaths %} active {%- endif -%}"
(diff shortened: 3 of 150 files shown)


linawolf (approved):
Great work, looks good to me!

linawolf:
@jaapio do you think we can backport this?

wouterj:
Just to understand it: is the main branch 2.x now and are we backporting things to 1.x to release a new minor? (cause Composer is configured that main is alias for 1.x as well)

linawolf:
@wouterj we are using version management similar to what TYPO3 does here. So we are always working on main and breaking changes are allowed on main. Whatever is non-breaking we can backport to 1.x. Releases for new minor or bugfix versions are done from 1.x. One day we hope main will become 2.0. The day that will happen we will branch of 2.0.

jaapio:
I never looked into details of branching as I do not maintain multiple versions, once I release a major version of a project I do normally just move on. I try to do no BC breaks and when I do so, I just move to the next major version.  That has worked for me for years. 

However this library is a bit different. There are multiple people actually contributing which is kinda new for me :-). So I just followed the model as suggested by @linawolf. I guess it's based on the gitflow model. The alias on main is an error from my side, that should be solved. 

Maybe we should have a look at this because right now mostly every pr we currently make is backported. We automated this, but still we need to think about it. So I'm open for suggestions to change this or other improvements. Let's move this discussion to an issue. 
---

I have no issues to merge this to v1 version, as the output format is not part of our BC promise as it is up to me. Backward compatibility for templates is a hard issue. Of course some changes are clearly not backward compatible, but as this PR is just focusing on whitespaces I do not see any issues with that.

jaapio (approved):
Thanks a lot @wouterj  for all these changes. It must have been a lot of work to check all the templates!

PR #1061, opened 2024-08-29
Fix invalid url parsing

jaapio:
I did not found any specification where an URL can contain parenthesis in the url, nor we do have a test that fails on this case. If we need it anyway we can introduce it later on. But for now this fixes a bug where the fetch of a url is way to gredy.


linawolf (approved):

linawolf:
Good solution!

PR #1078, opened 2024-09-24
Use a more precise autoload prefix

stof:
As `phpdocumentor/guides-markdown` provides only classes for the `Markdown` subnamespace, using a more precise autoload prefix for the PSR-4 configuration will allow the ClassLoader to avoid checking that package for other guides classes.
This won't make much difference when using an optimized autoloader with level 2 optimizations as the classmap will be used instead of checking the filesystem. But it will help a bit when using a non-optimized autoloader (which is common in dev environments as that provide a better DX when your codebase gets edited).

If this is considered good by maintainers, I will update the PR to apply that to other packages rather than just `guides-markdown` as others could also benefit from it.
Also, maintainers should decide whether they prefer using `src/Markdown/` as path in the config or flattening the `src` folder by removing the `Markdown` folder entirely (similar to what is done for `guides-cli`). Both options are valid and I'll happily apply the preferred one when updating this PR to cover all packages.

### DIFF
--- a/packages/guides-markdown/composer.json
+++ b/packages/guides-markdown/composer.json
@@ -9,7 +9,7 @@
     },
     "autoload": {
         "psr-4": {
-            "phpDocumentor\\Guides\\": "src/"
+            "phpDocumentor\\Guides\\Markdown\\": "src/Markdown/"
         }
     },
     "autoload-dev": {


jaapio:
Thanks for reaching out to us, I think it would be nice to have this optimization in the composer.json files of the projects. A PR for those changes would be very nice. 

Regarding the folder structure, I think we prefer the flattest structure, we must have missed that when we were splitting the project into separate packages. However changing this could have the downside of a harder to track history. I'm not sure if we should do that? @linawolf  what do you think?

linawolf:
I think if we are moving the files in their own commit without changing them further in that commit git should be able to realize they were moved and not deleted and then created a new

jaapio:
Ok, let's go for it @stof

linawolf:
Thanks for the PR and for the offer to extend it, @stof -- and sorry this sat without a response for so long.

We're going to close this and pick up the broader change ourselves in #1375: flattening src/<Namespace>/ into src/ and using a precise PSR-4 prefix across the packages where that's a clean fit, per the direction agreed above.

PR #1083, opened 2024-09-29
Reword paragraph

greg0ire:
Because it was a single sentence without punctuation, it was hard to
grasp that implementing the PrependExtensionInterface was optional.

Note that the lines below mention line numbers, but the line numbers are not shown on https://docs.phpdoc.org/components/guides/guides/developers/extensions/structure.html .


linawolf (approved):

linawolf:
Thanks for improving our docs

PR #1101, opened 2024-10-01
[BUGFIX] add support for auto detected inline links

### LINKED ISSUE #1096: [Bug]: Embedded URLS are not auto-detected in Markdown
### Summary

Embedded URLS are output as plain text

# Sample Markdown Document

See https://www.example.com for additional information.

See also https://www.example.com/more.

### Expected output

Embedded URLs are transformed into links

jaapio:
This fix adds the required extension to support inline links to be converted automatically to links.

fixes #1096


linawolf (approved):
Cool! Now we also know how to install an extension!

Issue #1086, opened 2024-09-30
[Bug]: Tables in Markdown are not displayed correctly


linawolf:
### Summary

See https://github.com/phpDocumentor/guides/pull/1085

### Expected output

See https://github.com/phpDocumentor/guides/pull/1085

linawolf:
There is an extension for that: https://commonmark.thephpleague.com/2.5/extensions/tables/

jaapio:
I added support for tables, but we do not have a system yet to set the alignment of a cell that can be done in markdown on a column level. I'm not sure we should add this feature.

linawolf:
I think if we need alignment support we can make a new issue. This one is solved for now

PR #1104, opened 2024-10-01
[FEATURE] add basic support for tables

jaapio:
This patch includes basic parsing of markdown tables. Tables in markdown seem to be very limited in perspective to what ReST supports. Missing feature right now is the cell allignment, which can be applied in markdown. But not in ReST.

refs #1086


linawolf:
Lets wait and see if anyone used that feature

PR #1113, opened 2024-10-02
[FEATURE] Handle image titles in Markdown

### LINKED ISSUE #1095: Handle image titles in Markdown
### Feature request

In Markdown it is possible to have an image with a title, for example:

```
![Example form2](Images/frontend2.png "Example Form with validation")
```

Currently the title gets lost in parsing and is not present in the node to be rendered

linawolf:
Resolves #1095

### DIFF
--- a/packages/guides/src/Nodes/Inline/ImageInlineNode.php
+++ b/packages/guides/src/Nodes/Inline/ImageInlineNode.php
@@ -20,8 +20,11 @@ final class ImageInlineNode extends InlineNode
 {
     public const TYPE = 'image';
 
-    public function __construct(private readonly string $url, private readonly string $altText)
-    {
+    public function __construct(
+        private readonly string $url,
+        private readonly string $altText,
+        private readonly string|null $title = null,
+    ) {
         parent::__construct(self::TYPE, $url);
     }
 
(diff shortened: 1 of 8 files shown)


packages/guides/src/Nodes/Inline/ImageInlineNode.php
```diff
@@ -20,8 +20,11 @@ final class ImageInlineNode extends InlineNode
 {
     public const TYPE = 'image';
 
-    public function __construct(private readonly string $url, private readonly string $altText)
-    {
+    public function __construct(
+        private readonly string $url,
+        private readonly string $altText,
+        private readonly string $title,
```
jaapio, line 26: Adding a new constructor parameter is a backwards compatible break. Please make this an optional value which is by default `null`

jaapio (changes requested):

packages/guides/src/Nodes/Inline/ImageInlineNode.php
linawolf: Done

linawolf:
@jaapio I made the argument optional, makes sense

jaapio (approved):

Issue #1108, opened 2024-10-01
Add support for automated menu creation


jaapio:
### Feature request

Markdown doesn't have anything like a `.. toctree::` in `ReST` This is not very nice as it requires a lot of hand work to create navigation. We want to provide an equal syntax as available for `ReST` to build navigation menu's

jaapio:
Possible related topics are:
https://commonmark.thephpleague.com/2.5/extensions/attributes/

jaapio:
because markdown doesn't have a real standard but about 13M flavors: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

jaapio:
https://docs.gitlab.com/ee/user/markdown.html#table-of-contents

jaapio:
https://rawgit.com/fletcher/MultiMarkdown-6-Syntax-Guide/master/index.html

PR #1116, opened 2024-10-02
[FEATURE] Make index file configurable

linawolf:
The index file will also be used in automatic menu building.

As oposed to setting "input-file" not only the one file mentioned will be rendererd but all files from the directory. However the name of the input file can be changed.

Some markdown projects like to change the name of the index file. For example powermail is using Readme.md as index file in every directory:

https://github.com/in2code-de/powermail/tree/master/Documentation

References #1108

### DIFF
--- a/packages/guides/src/Handlers/ParseDirectoryHandler.php
+++ b/packages/guides/src/Handlers/ParseDirectoryHandler.php
@@ -21,20 +21,28 @@ use phpDocumentor\Guides\Event\PostParseProcess;
 use phpDocumentor\Guides\Event\PreParseProcess;
 use phpDocumentor\Guides\FileCollector;
 use phpDocumentor\Guides\Nodes\DocumentNode;
+use phpDocumentor\Guides\Settings\ProjectSettings;
+use phpDocumentor\Guides\Settings\SettingsManager;
 use Psr\EventDispatcher\EventDispatcherInterface;
 
+use function array_map;
 use function assert;
+use function explode;
+use function implode;
 use function sprintf;
 
 final class ParseDirectoryHandler
 {
-    private const INDEX_FILE_NAMES = ['index', 'Index'];
+    private SettingsManager $settingsManager;
 
     public function __construct(
         private readonly FileCollector $fileCollector,
         private readonly CommandBus $commandBus,
         private readonly EventDispatcherInterface $eventDispatcher,
+        SettingsManager|null $settingsManager = null,
     ) {
+        // if for backward compatibility reasons no settings manager was passed, use the defaults
+        $this->settingsManager = $settingsManager ?? new SettingsManager(new ProjectSettings());
     }
 
     /** @return DocumentNode[] */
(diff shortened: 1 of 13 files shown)


packages/guides/src/Handlers/ParseDirectoryHandler.php
```diff
@@ -98,14 +101,23 @@ private function getDirectoryIndexFile(
             $hashedContentFromFilesystem[$itemFromFilesystem['basename']] = true;
         }
 
-        foreach (self::INDEX_FILE_NAMES as $indexName) {
+        if ($indexName !== null) {
             $indexFilename = sprintf('%s.%s', $indexName, $extension);
             if (isset($hashedContentFromFilesystem[$indexFilename])) {
                 return $indexName;
             }
```
jaapio, line 108: If you extract this part to a method we reduce the duplicated lines. As you repeated this in the foreach loop.

packages/guides/src/Handlers/ParseDirectoryHandler.php
```diff
@@ -98,14 +101,23 @@ private function getDirectoryIndexFile(
             $hashedContentFromFilesystem[$itemFromFilesystem['basename']] = true;
         }
 
-        foreach (self::INDEX_FILE_NAMES as $indexName) {
+        if ($indexName !== null) {
```
jaapio, line 104: if the user configured an `indexName`  we can enforce the index to exist like this. So I would expect a throw in this if block.

packages/guides/src/Handlers/ParseDirectoryHandler.php
```diff
@@ -98,14 +101,23 @@ private function getDirectoryIndexFile(
...
             }
+        } else {
+            foreach (self::INDEX_FILE_NAMES as $indexName) {
+                $indexFilename = sprintf('%s.%s', $indexName, $extension);
+                if (isset($hashedContentFromFilesystem[$indexFilename])) {
+                    return $indexName;
+                }
+            }
+
+            $indexName = self::INDEX_FILE_NAMES[0];
```
jaapio, line 117: This seems to be invalid to me... you fetch the first value, but we tried more. Why not include all valid option in the exception?

jaapio (changes requested):
I had some issues to follow the code in the parseDirectoryHandler. It looks like we can simplify it when we would use the settings manager, and skip the inline defaults? I think it would make sense to define a default in the settings manager. And remove the hardcoded defaults. 

Please keep in mind that adding a parameter to any class constructor should be seen as a backward compatibility break. So we might need to make them default to `null` when possible. And apply a fallback. Which would be in this case the inline defaults. `'index,Index'`

Given the fact that the inline defaults may contain multiple filenames. `index` and `Index` I think we should allow the users to specify multiple as well?

linawolf:
Is adding an additional constructor to a final class that is used by DI really breaking?

packages/guides/src/Handlers/ParseDirectoryHandler.php
linawolf: Good point, included them all now

packages/guides/src/Handlers/ParseDirectoryHandler.php
linawolf: We only have one for each loop now

linawolf:
@jaapio, tried to apply the changes as far as I understood them. Please have another look

packages/guides/src/Handlers/ParseDirectoryHandler.php
linawolf: I don't really get what you are suggesting?

jaapio (approved):

PR #1121, opened 2024-10-03
[FEATURE] Add option to create an automatic menu

linawolf:
For now all Document entries are added to the root document entry. In a follow up I plan to create a nested menu as well

References #1108


linawolf:
@jaapio tocnodes are replaced or removed now with warning

jaapio (approved):

PR #1128, opened 2024-10-11
[FEATURE] Support automatically created multi-level menus

linawolf:
Resolves https://github.com/phpDocumentor/guides/issues/1108


packages/guides/src/Compiler/Passes/AutomaticMenuPass.php
```diff
@@ -53,9 +64,46 @@ public function run(array $documents, CompilerContextInterface $compilerContext)
...
+            if (in_array($fileName, $indexNames, true)) {
+                array_pop($pathParts);
+                $path = implode('/', $pathParts);
+            }
+
+            $parentFound = false;
+            foreach ($indexNames as $indexName) {
+                $indexFile = $path . '/' . $indexName;
+                try {
+                    $parentEntry = $projectNode->getDocumentEntry($indexFile);
```
jaapio, line 87:
```suggestion
                    $parentEntry = $projectNode->findDocumentEntry($indexFile);
                    if ($parentEntry === null) {
                       continue;
                    }
```
This seems better to me as you are not using exceptions for situations that are not exceptional

packages/guides/src/Compiler/Passes/AutomaticMenuPass.php
```diff
@@ -53,9 +64,46 @@ public function run(array $documents, CompilerContextInterface $compilerContext)
...
+                continue;
+            }
+
+            try {
+                $parentEntry = $projectNode->getDocumentEntry($path);
+                $documentEntry->setParent($parentEntry);
+                $parentEntry->addChild($documentEntry);
+            } catch (DocumentEntryNotFound) {
+                $this->logger?->warning(sprintf('No parent found for file "%s/%s" attaching it to the document root instead. ', $path, $fileName));
+            }
```
jaapio, line 106: Same here please use `findDocumentEntry`

jaapio (changes requested):
Some minor changes that will improve the speed

packages/guides/src/Compiler/Passes/AutomaticMenuPass.php
linawolf: Done

packages/guides/src/Compiler/Passes/AutomaticMenuPass.php
linawolf: Done

jaapio (approved):

PR #1133, opened 2024-10-11
[FIX] use canonicalUrl to fetch images

jaapio:
This will solve the situations containing relative urls.


linawolf:
I think a test for that would be nice?

linawolf:
I added a test and also tried it out with the powermail guide: works fine and code looks good

linawolf (approved):

jaapio:
Thanks for adding the test

Issue #1160, opened 2024-10-22
:bug:  Nested inline nodes are not supported


jaapio:
### Summary

Markdown does support nested styling like _**bold italic**_  right now this is not allowed within this project. But also multiple plaintextnodes are not supported 

A simple fix for multiple plaintext nodes could be:

`guides-markdown/src/Markdown/Parsers/InlineParsers/AbstractInlineTextDecoratorParser.php`
```
                $plainText = '';
                $showWarning = false;
                foreach ($content as $node) {
                    if ($node instanceof PlainTextInlineNode === false) {
                        $showWarning = true;
                    }

                    $plainText .= $node->getValue();
                }

                if ($showWarning) {
                    $this->logger->warning(sprintf('%s CONTEXT: Content of emphasis could not be interpreted: %s', $this->getType(), var_export($content, true)));
                }

                return $this->createInlineNode($commonMarkNode, $plainText);
```

### Code snippet that reproduces the problem

```
 _**bold italic**_ 
```

### Expected output

```
<i><b>bold italic</b></i>
```

Issue #1161, opened 2024-10-25
Deprecated use of string values for inline compound nodes


jaapio:
### Feature request

Inline nodes extending `InlineCompoundNode` should not be accessed via `getValue` but `getChildren` should be used in the templates. 

```
    {%- for child in node.children -%}
        {{- renderNode(child) -}}
    {%- endfor -%}
```

We will remove the backward compatibility layer in version 2.

PR #1162, opened 2024-10-25
Allow some inlinenodes to have child nodes

jaapio:
Markdown supports more nesting on nodes than RST does. To support this we needed to add support for compound nodes. We tried to be backward compatible for people building their own template. This layer will be removed in v2 and therfor we do trigger deprecations. See #1161

Fixes: #1160

### DIFF
--- a/packages/guides/src/Nodes/Inline/AbstractLinkInlineNode.php
+++ b/packages/guides/src/Nodes/Inline/AbstractLinkInlineNode.php
@@ -13,13 +13,32 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Guides\Nodes\Inline;
 
-abstract class AbstractLinkInlineNode extends InlineNode implements LinkInlineNode
+use Doctrine\Deprecations\Deprecation;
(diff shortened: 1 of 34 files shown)


packages/guides/src/Nodes/Inline/AbstractLinkInlineNode.php
```diff
@@ -13,13 +13,32 @@
 
 namespace phpDocumentor\Guides\Nodes\Inline;
 
-abstract class AbstractLinkInlineNode extends InlineNode implements LinkInlineNode
+use Doctrine\Deprecations\Deprecation;
+use phpDocumentor\Guides\Nodes\InlineCompoundNode;
+
+abstract class AbstractLinkInlineNode extends InlineCompoundNode implements LinkInlineNode
```
wouterj, line 19: Just a fyi for next time: the change of parent class from `InlineNode` to `InlineCompoundNode` is a BC break for people that extend this base class in their code.

I've fixed this now in Symfony's alpha code, but it might be worth a notion in the release notes (and generally, I would advice to not introduce new deprecations in patch versions).

PR #1184, opened 2024-11-30
Change inline node deprecation and fix deprecations in the codebase

wouterj:
After upgrading to the latest guide for the first time since a long time, I saw quite some deprecations of #1162 coming from Guides itself. This PR enables deprecations in the testsuite, so that any deprecated usage in the library is noticed and fixed.

While working on that, I discovered we can take #1162 one step further. That PR adds a `array $children` parameter to some inline nodes that must be used instead of the existing `$value` parameter, but it doesn't actually deprecate the `$value` parameter.
This PR:
- Replaces the `$value` parameter with the new `$children` parameter (in a BC matter)
- Deprecates the old `$children` parameter location

At last, enabling deprecations showed a couple other deprecations that I've fixed as well to make sure the tests are green.

---

Sorry for the slightly chaotic PR contents. I hope it all makes sense now and happy to elaborate further if needed.


phpunit.xml.dist
greg0ire, line 1: Very interesting change, I should try to do the same on Doctrine itself!

packages/guides/tests/unit/Compiler/CompilerContextTest.php
```diff
@@ -15,19 +15,22 @@
...
 use Doctrine\Deprecations\PHPUnit\VerifyDeprecations;
 use phpDocumentor\Guides\Nodes\ProjectNode;
+use PHPUnit\Framework\Attributes\WithoutErrorHandler;
 use PHPUnit\Framework\TestCase;
 
 final class CompilerContextTest extends TestCase
 {
     use VerifyDeprecations;
 
+    #[WithoutErrorHandler]
```
greg0ire, line 25: Why is this necessary? The test passes when I remove it.

wouterj: see https://github.com/doctrine/deprecations/pull/66#issuecomment-2509771433

packages/guides-theme-rst/resources/template/rst/inline/emphasis.rst.twig
```diff
@@ -1 +1 @@
-*{{- node.value|raw -}}*
+*{{- node|plaintext -}}*
```
jaapio, line 1: When I introduced the inline nested nodes I did use the value because that would not require a change in the templates. As templates can be overwritten by consuming projects. I did see that as a backward compatibility issue. 

This was the reason to try to be backwards compatible and accept the issues in our pipeline. How do you see that? Bc in templates is a very complex problem as I experienced in the past. But maybe I will try to be too strict about this project?

jaapio (commented):
Thanks a lot for your changes. It's good to see that you are still trying to align with us. I do the improvements you did. But I do have a question about how you look at backward compatibility, and templates.

packages/guides-theme-rst/resources/template/rst/inline/emphasis.rst.twig
wouterj: I'm not sure I can follow you. The original content still works, but triggers a deprecation. Are you thinking of a case where the user overrides `getValue()` of a node that now has a BC break because it isn't called anymore?

You can use `Deprecation::triggerIfCalledFromOutside()` in `getValue()` then (or the Symfony tactic: at a virtual new boolean parameter to `getValue()` that determines whether the deprecation should be triggered). This should not produce a deprecation when called from a Guides template, but does when called by a user.

In general, deprecations are most useful if all of them can be fixed by a user without upgrading the package triggering the deprecation. If Guides triggers a deprecation from itself, I would never know if I can safely upgrade to 2.0.

PR #1198, opened 2025-01-21
Upgrade flysystem v2

jaapio:
This is an effort to abstract away the logic provided by FlySystem. FlySystem has some challenges when updating, but also limits the way filesystems are implemented in this project. By adding a new package we can migrate to a new version of flysystem, and even introduce other file system adapters when people want to use a different system. Although the latter requires some extra work to allow people to use the flysystem based file finder. 

The new system has been implemented in a backward compatible way, so other projects depending on this project can migrate when they are ready.


jaapio:
@linawolf, I would like to hear what you think about these changes, **please do NOT update the branch** as a lot of composer steps have been taken by me. I will take care of merging when we agree this is a path we should take. 

I think this path will also allow us to get files outside the documentation directory for typo3 projects, by replacing flysystem with another filesystem like symfony/filesystem for example.

PR #1231, opened 2025-06-17
First iteration on option attributes

jaapio:
## Directive attributes & compile-time directive processing

Introduces a new attribute-based declaration system for reStructuredText directives and moves directive processing from the parse phase into the compile phase.

Moving directive processing to a compiler pass is the groundwork that enables upcoming features such as option validation — with directives now declaratively describing their options via attributes, the framework can inspect and validate them centrally instead of each directive doing ad-hoc parsing.

### Features

**`#[Directive]` and `#[Option]` attributes**
- New `phpDocumentor\Guides\RestructuredText\Directives\Attributes\Directive` attribute declares a directive's name and aliases on the class, replacing the abstract `getName()` / `getAliases()` methods.
- New repeatable `#[Option]` attribute (`OptionType`, default value, description, example) declares the options a directive supports — used for automatic option validation, type coercion, and documentation.

**Compile-time directive processing**
- New `DirectiveNode` represents a parsed directive during compilation and is now processed by a dedicated `DirectiveProcessPass` compiler pass instead of being resolved eagerly by the parser.
- New `ReverseNodeTransformer` interface, with support in `DocumentNodeTraverser`, so parent transforms run after their children.
- Directives expose a `createNode(DirectiveNode)` entry point; a backward-compat `process()` path (with `isUpgraded()`) keeps existing directives working while they are migrated.

**Upgraded directives**
- Migrated to the attribute model: admonitions (attention, caution, danger, error, hint, important, note, tip, warning), version-change directives (versionadded, versionchanged, deprecated), figure, image, include, confval, youtube, and others.
- `readOption()` / `readAllOptions()` on `BaseDirective` now coerce values to the declared `OptionType` and apply attribute defaults (e.g. youtube width/height/allow/allowfullscreen).
- Admonitions now correctly apply the option-provided CSS `class` to the rendered node.

### Deprecations
- Directives without a `#[Directive]` attribute now trigger a deprecation when their name is resolved, pointing to documentation on migrating custom directives.

### Notes
- Includes a base-line merge of `update-security-issues` and a dependency bump.
- This is a first iteration; remaining built-in directives will be migrated to the attribute/compile-time model in follow-ups.

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Directives/BaseDirective.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Directives/BaseDirective.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Guides\RestructuredText\Directives;
 
+use Doctrine\Deprecations\Deprecation;
 use LogicException;
 use phpDocumentor\Guides\Nodes\GenericNode;
 use phpDocumentor\Guides\Nodes\Node;
@@ -110,7 +111,18 @@ abstract class BaseDirective
         $reflection = new ReflectionClass($this);
         $attributes = $reflection->getAttributes(Attributes\Directive::class);
 
-        return count($attributes) === 1;
+        if (count($attributes) !== 1) {
+            Deprecation::trigger(
+                'phpdocumentor/guide-restructured-text',
+                'https://github.com/phpDocumentor/guides/issues/1373',
+                'Directive class "%s" must have a Directive attribute. This directive needs to be upgraded.',
+                $this->getName(),
+            );
+
+            return false;
+        }
+
+        return true;
     }
 
     /**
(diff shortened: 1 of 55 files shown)


packages/guides-restructured-text/src/RestructuredText/Directives/BaseDirective.php
```diff
@@ -34,10 +45,34 @@
...
+        if (isset($this->name)) {
+            return $this->name;
+        }
+
+        $reflection = new ReflectionClass($this);
+        $attributes = $reflection->getAttributes(Attributes\Directive::class);
+
+        if (count($attributes) === 0) {
+            throw new LogicException('Directive class must have a Directive attribute');
+        }
```
wouterj, line 70: Unless I'm wrong, this is a BC break. This should be a deprecation trigger in 1.x, and only become an exception in 2.0 (at least, all of our test suite is now failing with this exception 😄)

jaapio: You should implement the getName, maybe the removal of abstract is confusing?

wouterj: This is the directive that is causing the trouble (together with the other one): https://github.com/wouterj/docs-builder/blob/phpdocumentor-guides/src/GuidesExtension/src/Directives/BestPracticeDirective.php

The problem is actually that `AbstractAdmonitionDirective` removed the `getName()` method. I'll write a patch. I have a plan to add a "bc check" test job to the CI, so I try to combine both in 1 PR.

PR #1233, opened 2025-06-24
[BUGFIX] Check error level is higher

jaapio:
When the error level is set to warning on the spy errors should also be found as fail.


linawolf (approved):
Cool, we can even reduce the phpstan baseline and you included tests.

PR #1242, opened 2025-08-20
Initial setup for file serve

jaapio:
With this change we provide an internal webserver in the application. And introduce a watch on changes. This introduces a new way of working with guides, allowing users to edit documentation without caring about rerendering.

Optimizations are possible like parsing and rendering a single file to make rerendering faster.


garvinhicking:
When editing a file on my macOS (Orbstack) I only get this:

```
string(24) "Unhandled event mask: 10"
```

and the file is not re-rendered.

PR #1248, opened 2025-09-20
Micro optimizations

jaapio:
- array_walk is slower than a foreach, so buffer now uses foreach.
- By caching some of the inline node parsers we can strip out about 9% of
  the overhead in the inline parser, as we do not have to check them for every token.


linawolf (approved):
This looks good to me, thank you!

PR #1269, opened 2025-11-28
Allow Symfony 7 and 8

wouterj:
Part of this change was done [before by Gregoire](https://github.com/phpDocumentor/guides/pull/793), but dependabot somehow override it back to only Symfony 6. Is there a way to make sure this does not happen again?

For integration on symfony.com, we would need at least Symfony 7 support at the moment. So this is a quick change to allow it. Later on, we probably would also want to run the test suite against Symfony 7 and 8 in GitHub Actions.


jaapio:
I don't know how to tell dependabot not to drop versions. I'm sure there must be an option but I never looked into this.

jaapio:
Thanks for this PR, I just merged #1270  because that one contained a valid pipeline. I will create a PR to make sure dependabot respects this range.

PR #1279, opened 2026-01-02
[BUGFIX] Ensure temp directory exists before calling tempnam() + DI support

CybotTM:
## Summary

The `tempnam()` call in `PlantumlRenderer::render()` uses a subdirectory of `sys_get_temp_dir()` that may not exist, which triggers a PHP `E_NOTICE`:

```
Notice: tempnam(): file created in the system's temporary directory
```

This notice can appear in rendered documentation output when error reporting includes `E_NOTICE`.

## Changes

- Create the temp subdirectory if it doesn't exist before calling `tempnam()`
- Add dependency injection for temp directory path (optional constructor parameter, defaults to `sys_get_temp_dir() . '/phpdocumentor'`)
- Add regression test using injected temp directory for proper isolation

## Test plan

- [ ] Configure guides to use local PlantUML binary (`renderer="plantuml"`)
- [ ] Ensure `/tmp/phpdocumentor` does not exist
- [ ] Render documentation with a `.. uml::` directive
- [ ] Verify no `E_NOTICE` is emitted

---

*Moved from https://github.com/phpDocumentor/guides-graphs/pull/2*

### DIFF
--- a/packages/guides-graphs/src/Graphs/Renderer/PlantumlRenderer.php
+++ b/packages/guides-graphs/src/Graphs/Renderer/PlantumlRenderer.php
@@ -28,10 +28,14 @@ use function tempnam;
 
 final class PlantumlRenderer implements DiagramRenderer
 {
-    public const TEMP_SUBDIRECTORY = '/phpdocumentor';
+    private readonly string $tempDirectory;
 
-    public function __construct(private readonly LoggerInterface $logger, private readonly string $plantUmlBinaryPath)
-    {
+    public function __construct(
+        private readonly LoggerInterface $logger,
+        private readonly string $plantUmlBinaryPath,
+        string|null $tempDirectory = null,
+    ) {
+        $this->tempDirectory = $tempDirectory ?? sys_get_temp_dir() . '/phpdocumentor';
     }
 
     public function render(RenderContext $renderContext, string $diagram): string|null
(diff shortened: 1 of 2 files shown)


packages/guides-graphs/src/Graphs/Renderer/PlantumlRenderer.php
```diff
@@ -45,7 +49,12 @@ public function render(RenderContext $renderContext, string $diagram): string|nu
 @enduml
 PUML;
 
-        $pumlFileLocation = tempnam(sys_get_temp_dir() . '/phpdocumentor', 'pu_');
+        $tempDir = sys_get_temp_dir() . self::TEMP_SUBDIRECTORY;
+        if (!is_dir($tempDir)) {
+            mkdir($tempDir, 0o777, true);
```
jaapio: I'm not sure we should create a 777 permission directory.  `0755` is sufficient and lowers the attack surface for possible exploits.

packages/guides-graphs/src/Graphs/Renderer/PlantumlRenderer.php
```diff
@@ -45,7 +49,12 @@ public function render(RenderContext $renderContext, string $diagram): string|nu
 @enduml
 PUML;
 
-        $pumlFileLocation = tempnam(sys_get_temp_dir() . '/phpdocumentor', 'pu_');
+        $tempDir = sys_get_temp_dir() . self::TEMP_SUBDIRECTORY;
```
jaapio: Maybe it's better to inject the directory used by this class, I remember when I wrote this I didn't write a test because it was hard to run this in isolation. Injecting the path can be done via the constructor and would allow us to use vfsStream to fake the filesystem.

CybotTM: Good suggestion for testability! This was initially out of scope for the bugfix, but I can include it if you'd like. Let me know.

jaapio: If you can do that, that would be great. 

Please keep in mind that we do not want to be backwards compatible. So any new constructor argument must have a default value. Which could still be sys_get_tmp_dir.

jaapio (approved):

PR #1280, opened 2026-01-02
[FEATURE] Add :nbsp: text role for non-breaking spaces

CybotTM:
## Summary

Implements the `:nbsp:` text role which produces a non-breaking space (`&nbsp;`). This provides an explicit alternative to the `~` syntax (e.g., `a~b`) for inserting non-breaking spaces.

**Usage:**
```rst
This is a non breakable space: a :nbsp:`ignored` b
```

The content inside the role is ignored; the role simply produces a non-breaking space.

## Background

- The functional test for `:nbsp:` was imported from doctrine/rst-parser in commit `77e1f40c`
- The infrastructure (`WhitespaceInlineNode` and `nbsp.html.twig` template) already existed
- Only the text role class was missing, so the test was marked as `SKIP`

## Changes

- Add `NbspTextRole` class that returns `WhitespaceInlineNode`
- Register the text role in the service configuration
- Remove `SKIP` marker from the functional test

## Test plan

- [x] Run functional test `nbsp-role` - now passes
- [x] PHPCS - no errors
- [x] PHPStan - no errors
- [x] Unit tests - all pass
- [x] Functional tests - 9 incomplete (was 10)

### DIFF
--- a/tests/Functional/tests/nbsp-role/nbsp-role.html
+++ b/tests/Functional/tests/nbsp-role/nbsp-role.html
@@ -1,3 +1,2 @@
-SKIP non-breacking space role does not work
 <p>This is a non breakable space: a&nbsp;b</p>
 <p>This is also a non breakable space: a &nbsp; b</p>


CybotTM:
## Alternative: Remove the test instead?

After implementing this, I'm questioning whether the `:nbsp:` role is actually needed:

**The `~` syntax already works:**
```rst
a~b  →  a&nbsp;b
```

**The `:nbsp:` role is more verbose and requires dummy content:**
```rst
a :nbsp:`x` b  →  a &nbsp; b
```

### Options

1. **Merge this PR** - Keep the role for feature parity with doctrine/rst-parser
2. **Remove the test instead** - The `~` syntax is sufficient, less code to maintain
3. **Keep as SKIP** - Leave it as a known limitation

### Question for maintainers

The test was imported from doctrine/rst-parser (commit `77e1f40c`), but the implementation was never added. Was this intentional (`:nbsp:` role not wanted) or an oversight?

If the `:nbsp:` role isn't needed, I'm happy to close this PR and submit one that removes the `tests/Functional/tests/nbsp-role/` directory instead.

packages/guides-restructured-text/src/RestructuredText/TextRoles/NbspTextRole.php
```diff
@@ -0,0 +1,39 @@
...
+use phpDocumentor\Guides\Nodes\Inline\WhitespaceInlineNode;
+use phpDocumentor\Guides\RestructuredText\Parser\DocumentParserContext;
+
+/**
+ * Role for inserting a non-breaking space.
+ *
+ * Usage: :nbsp:`ignored content`
+ *
+ * The content is ignored; the role simply produces a non-breaking space.
+ * This is an alternative to the ~ syntax (e.g., a~b).
```
jaapio: `This is an alternative to the ~ syntax (e.g., a~b).` 

This comment seems to be invalid, the `~` syntax is not supported, and also not part of the ReStructuredText spec. Therefore I think we should remove this comment to avoid confusion by our users.

jaapio (approved):

PR #1289, opened 2026-01-22
perf: Add DI container caching for faster CLI startup

CybotTM:
## Summary

Caches the compiled Symfony DI container to `/tmp/guides-container-cache/` for faster CLI startup times.

### Changes

- **ContainerFactory**: Cache compiled container using PhpDumper
  - Generates unique cache key from vendor directory path
  - Stores compiled container as PHP class file
  - Automatic invalidation when container configuration changes
  - Graceful fallback when container has uncacheable parameters

### Performance Impact

Container compilation is skipped on subsequent runs, reducing startup time. Most beneficial for development workflows with frequent CLI invocations.

See [Performance Analysis Report](https://cybottm.github.io/render-guides/) for detailed benchmarks.

---

## Related PRs

| PR | Description | Status |
|----|-------------|--------|
| #1287 | Rendering caching layer | Independent |
| #1288 | RST parsing optimizations | Independent |
| #1289 | **This PR** - CLI container caching | |
| #1291 | Symfony 8 compatibility | ✅ Merged |
| #1293 | ProjectNode O(1) document lookup | Independent |

All PRs can be merged independently in any order.

### DIFF
--- a/packages/guides-cli/src/DependencyInjection/ContainerFactory.php
+++ b/packages/guides-cli/src/DependencyInjection/ContainerFactory.php
@@ -22,20 +22,35 @@ use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Dumper\PhpDumper;
+use Symfony\Component\DependencyInjection\Exception\RuntimeException as DIRuntimeException;
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 
+use function array_keys;
 use function array_merge;
+use function assert;
 use function class_exists;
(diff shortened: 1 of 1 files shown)


CybotTM:
Closing this: the cache is never written, so the change delivers no speedup at all. `GuidesExtension` attaches a `ProjectSettings` object via `addMethodCall('setProjectSettings', ...)`, and `PhpDumper::dump()` refuses any container that holds an object parameter; the resulting `RuntimeException` is swallowed by the `catch (DIRuntimeException)`. Measured on the current head: after a full render `/tmp/guides-container-cache` exists and is empty, and container creation takes the same time as on `main`.

Independently of that, the read path `require_once`s PHP from a hardcoded world-writable `/tmp` directory, which I would not want to ship even once the write path works. Making the container dumpable — moving `ProjectSettings` into a service definition — is a separate and larger change; if I pick this up again it will be a new pull request on that basis.

PR #1295, opened 2026-01-23
feat: Add PSR-16 caching support to JsonLoader

CybotTM:
## Summary

Adds PSR-16 caching to `JsonLoader` for inventory JSON files. By default, uses an in-memory `ArrayAdapter` that deduplicates HTTP requests within a process. For persistent caching across CLI invocations, inject any PSR-16 compatible cache.

## Changes

```php
// Constructor signature
public function __construct(
    private readonly HttpClientInterface $client,
    CacheInterface|null $cache = null,  // PSR-16 cache (default: ArrayAdapter)
)
```

## Behavior

| `$cache` value | Behavior |
|----------------|----------|
| `null` (default) | In-memory ArrayAdapter - deduplicates within same process |
| `FilesystemAdapter` | Persistent caching across CLI invocations |
| `ChainAdapter` | Multi-tier caching (memory + filesystem) |

## Features

- **PSR-16 interface**: Use any compatible cache implementation
- **In-memory default**: Automatic request deduplication within same process
- **Multi-tier support**: Use Symfony ChainAdapter for L1 memory + L2 filesystem
- **Hash-based keys**: Uses xxh128 for URL → cache key mapping

## Performance Impact

~53% faster rendering for interlink-heavy documentation by avoiding repeated HTTP requests for inventory files.

## Backward Compatibility

This is fully backward compatible - new optional parameter with sensible default.

## Usage Examples

```php
// Default - in-memory deduplication
$loader = new JsonLoader($httpClient);

// Persistent filesystem cache (TTL configured on adapter)
$pool = new FilesystemAdapter('inventory', 3600, '/tmp/cache');
$loader = new JsonLoader($httpClient, new Psr16Cache($pool));

// Multi-tier: memory + filesystem
$chain = new ChainAdapter([
    new ArrayAdapter(),
    new FilesystemAdapter('inventory', 3600, '/tmp/cache'),
]);
$loader = new JsonLoader($httpClient, new Psr16Cache($chain));
```

## Test Plan

- [x] All existing tests pass
- [x] New tests for cache hit/miss, deduplication, deterministic keys
- [x] PHPStan passes
- [ ] CI validation

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1715,7 +1715,7 @@
             "dist": {
                 "type": "path",
                 "url": "./packages/guides",
-                "reference": "569c35aee270c467111959c7e4311d1d13eb82df"
+                "reference": "679fbae8e696ec5007fa3ccf26148874949c5854"
             },
             "require": {
                 "doctrine/deprecations": "^1.1",
@@ -1728,6 +1728,8 @@
                 "phpdocumentor/filesystem": "^1.7.4",
                 "phpdocumentor/flyfinder": "^1.1 || ^2.0",
                 "psr/event-dispatcher": "^1.0",
+                "psr/simple-cache": "^3.0",
+                "symfony/cache": "^6.4.8 || ^7.4 || ^8",
                 "symfony/clock": "^6.4.8 || ^7.4 || ^8",
                 "symfony/html-sanitizer": "^6.4.8 || ^7.4 || ^8",
                 "symfony/http-client": "^6.4.9 || ^7.4 || ^8",
@@ -2065,6 +2067,55 @@
             }
         },
         {
+            "name": "psr/cache",
+            "version": "3.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/php-fig/cache.git",
+                "reference": "aa5030cfa5405eccfdcb1083ce040c2cb8d253bf"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/php-fig/cache/zipball/aa5030cfa5405eccfdcb1083ce040c2cb8d253bf",
+                "reference": "aa5030cfa5405eccfdcb1083ce040c2cb8d253bf",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=8.0.0"
+            },
+            "type": "library",
+            "extra": {
+                "branch-alias": {
+                    "dev-master": "1.0.x-dev"
+                }
+            },
+            "autoload": {
+                "psr-4": {
+                    "Psr\\Cache\\": "src/"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "authors": [
+                {
+                    "name": "PHP-FIG",
+                    "homepage": "https://www.php-fig.org/"
+                }
+            ],
+            "description": "Common interface for caching libraries",
+            "keywords": [
+                "cache",
+                "psr",
+                "psr-6"
+            ],
+            "support": {
+                "source": "https://github.com/php-fig/cache/tree/3.0.0"
+            },
+            "time": "2021-02-03T23:26:27+00:00"
+        },
+        {
             "name": "psr/clock",
             "version": "1.0.0",
             "source": {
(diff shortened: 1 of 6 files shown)


jaapio (changes requested):
This idea has been on my list for a while. Thanks for this improvement. 

I would like to see usage of a psr interface for the cache storage. This could be either psr 6 or 16. This removes the file dependencies from this implementation, and makes it more flexible for different purposes.

 We could use symfony cache by default in the cli application. To provide a simple usage for the people using this as a tool. While people using the library can still chose their preferred implementation. Maybe we could even use an in memory adapter by default?

jaapio (changes requested):
I notice that the number of PRs and the size of them are taking a lot of time. I do appreciate all the work you spend, but please have a look at the content you are pushing. 

I'm trying to read all you comments, and posts but some of them are so obvious AI generated that I'm starting to ignore your messages.  I don't think that's worth my time, but also not yours. So I would like to ask you to carefully review yourself what ever is posted. Otherwise I will have to close the PR without comments.

CybotTM:
Hi @jaapio,

I looked at this again after some distance. I think I understand better now where the discussion went wrong.

I used the PR to make the performance problem concrete, because this kind of change is easier to discuss with working code. I still think that can be useful. But I also see that this PR grew into more than a small discussion artifact: PSR-16 integration, default cache behavior, Symfony cache dependency, documentation, tests, and changed error messages. That creates real review and maintenance cost.

I do not see the PR as a gift that creates an obligation to merge it. The relevant point from the article is clear: even useful contributions still cost maintainer time, especially when they affect architecture and long-term ownership.

The problem I wanted to solve remains:

- interlink-heavy documentation repeatedly fetches the same inventory JSON
- this can make local rendering and CI significantly slower
- my local version improves that use case substantially

But the project-level design should be agreed first:

- should `JsonLoader` know about caching at all?
- should caching be injected there, or configured higher in the CLI/application layer?
- should the library depend only on `psr/simple-cache`, or also on `symfony/cache` for a default implementation?
- should the default be no cache, in-memory cache, or CLI-configured cache only?
- what documentation is useful without becoming noise?

About the longer AI-assisted comments: I want to explain my intention, because I think this may have created the wrong impression.

English is not my strongest language. AI helps me express technical points more clearly and unambiguously, and it reduces the risk that my wording sounds harsher or less precise than intended. I also tend to value more context in PRs, because many short PRs become difficult to understand later when the original intention is no longer obvious.

But I understand the other side as well: if the review thread becomes too verbose, it increases your workload instead of helping. I can handle that better by keeping review comments shorter and moving the deeper rationale into the PR description or a separate design note.

My intention is not to generate noise. My intention is to communicate clearly despite the language barrier and to leave enough context for future maintainers.

If you are still open to the feature, I can reduce this to the smallest architectural proposal first, without adding more implementation detail before the direction is agreed.

jaapio:
Hi, 

Thanks for getting back to this. I think it's still useful to introduce caching for the inventories. Since you created this PR I also refactored some parts of the Interlink resolving. I introduced some new concept which allows us to have multiple Resolvers. I added some basic documentation on how to add a custom resolver. https://docs.phpdoc.org/components/guides/guides/developers/extensions/interlinks.html, this is not directly relevant for what you want to achieve, but I wanted to point this out as it might have impact on your changes. 

When we implemented the JsonLoader I was thinking that the caching could be done in the http client that is injected into the loader. Like a middleware for http caching. When I look at the implementation again that we did back then I think it would make more sense deprecate the current one. Maybe the same applies to the DefaultInventoryRepository. 

My idea is that the new ChainedInventoryResolver gets multiple InventoryLinkResolvers. An InventoryLinkResolver is responsible to load the remote file and resolve the link from the file. When `InventoryLinkResolver::resolveInventoryLink` is called it should do what ever is needed to resolve the link or return a null when nothing is available.  Maybe we can introduce some class named "PSRInventoryResolver" that gets the logic to use a PSR-18 http client to do a http request to load an external inventory file. The External file could be cached by the `PSRInventoryResolver` in a property. This introduced a simple caching system during a single run. We could add a second constructor argument which can be a PRS-16 cache interface. If we make this optional, users of phpDocumentor/guides can decide to use caching or not. 

In phpDocumentor/guides-cli we implement the usage of this new class, chosing a default cache and http client. 

By making this separation we allow users of phpDocumentor/guides to choose there own configuration while CLI users like you get the default behavior out of the box.

We can mark all classes that are not required anymore with this new approach deprecated, and remove them on v2 of guides. 

I hope this idea is a bit clear and answers your questions about where to implement things. If I need to explain more or when you have different idea's I would love to hear those.

CybotTM:
Hi @jaapio,

sounds much better than my JsonLoader "bloating".

Maybe the name `PSRInventoryResolver` could be improved by `HttpInventoryResolver` or something similar.

So IMO we should close this PR.

PR #1296, opened 2026-01-23
feat: Add incremental build infrastructure for dependency tracking

CybotTM:
## Summary

Adds core infrastructure for incremental documentation builds, enabling consumers to:

- **Track document dependencies** - Know which documents import others via toctree, doc refs, and includes
- **Detect changes efficiently** - Fast mtime-first detection with content hash verification
- **Propagate invalidation** - When a document changes, find all dependents that need re-rendering
- **Persist state** - Serialize/deserialize build state between runs

### Core Components

| Component | Purpose |
|-----------|---------|
| `DependencyGraph` | Bidirectional graph tracking import/dependent relationships with cycle detection |
| `DocumentExports` | Immutable container for document anchors, titles, citations |
| `ChangeDetector` | Fast change detection using mtime + content hash strategy |
| `ContentHasher` | File hashing with xxh128 (fast) or sha256 fallback |
| `IncrementalBuildState` | Container service holding graph + exports during compilation |
| `DependencyGraphPass` | Compiler pass extracting imports from document AST |
| `ExportsCollectorPass` | Compiler pass collecting exports after compilation |

### Orchestration Layer

| Component | Purpose |
|-----------|---------|
| `IncrementalBuildCache` | Main cache orchestrator with sharded storage (256 buckets using MD5 hash) |
| `DirtyPropagator` | Propagates dirty state through dependency graph with export comparison |
| `GlobalInvalidationDetector` | Detects when full rebuild is needed (config/theme/toctree changes) |
| `PropagationResult` | DTO for dirty propagation results |
| `CacheVersioning` | Cache version validation (PHP version, format version) |

### Security Hardening

- **Resource exhaustion protection**: Limits on documents (100k), edges (2M), exports (100k), imports per doc (1k)
- **Path traversal prevention**: realpath() validation with prefix checking
- **Input validation**: Comprehensive validation in all `fromArray()` deserialization
- **Depth limiting**: Node traversal limited to 100 levels to prevent stack overflow
- **Propagation limits**: MAX_PROPAGATION_VISITS (100k) prevents runaway propagation

### Documentation

- `incremental-builds.rst` - Consumer guide with usage examples and integration patterns
- `incremental-builds-architecture.rst` - Maintainer guide with architecture, security model, and extension instructions

### Usage

Consumer applications integrate by:

1. Loading previous `IncrementalBuildCache` from cache (if exists)
2. Running compilation (passes populate the state automatically)
3. Using `ChangeDetector` to determine which documents need re-rendering
4. Using `DirtyPropagator::propagate()` to find affected dependents (with export comparison)
5. Using `GlobalInvalidationDetector` to check if full rebuild is needed
6. Persisting state via `IncrementalBuildCache::save()` for next build

See `docs/developers/incremental-builds.rst` for detailed documentation.

## Test Plan

- [x] 212 unit tests covering all components
- [x] Security tests for path traversal, symlink attacks, control characters
- [x] Boundary tests for resource limits (100k docs, 2M edges, 100k exports)
- [x] Stress tests for large graph operations (linear chains, fan-out patterns)
- [x] Pattern validation tests for GlobalInvalidationDetector
- [x] TOCTOU race condition handling documented and tested
- [x] PHPStan clean
- [x] Coding standards clean

### DIFF
--- a/docs/developers/incremental-builds-architecture.rst
+++ b/docs/developers/incremental-builds-architecture.rst
@@ -1,0 +1,304 @@
+====================================
+Incremental Builds: Architecture
+====================================
+
+This document describes the internal architecture, design decisions, and
+security considerations of the incremental build system. For usage documentation,
+see :doc:`incremental-builds`.
+
+Design Goals
+============
+
+The incremental build system was designed with these priorities:
+
+1. **Correctness** - Never skip a document that needs re-rendering
+2. **Performance** - O(1) operations where possible, efficient memory usage
+3. **Security** - Prevent resource exhaustion and path traversal attacks
+4. **Parallelization** - Support parallel compilation workflows
+
+Architecture Overview
+=====================
+
+.. code-block:: text
+
+    ┌─────────────────────────────────────────────────────────────┐
+    │                    IncrementalBuildCache                     │
+    │  (Orchestrates caching, persistence, and state management)   │
+    └─────────────────────────────────────────────────────────────┘
+                │                    │                    │
+                ▼                    ▼                    ▼
+    ┌───────────────────┐ ┌──────────────────┐ ┌─────────────────┐
+    │  DependencyGraph  │ │  DocumentExports │ │  CacheVersioning│
+    │  (Import/export   │ │  (Per-document   │ │  (Version       │
+    │   relationships)  │ │   public API)    │ │   validation)   │
+    └───────────────────┘ └──────────────────┘ └─────────────────┘
+                │                    │
+                ▼                    ▼
+    ┌───────────────────┐ ┌──────────────────┐
+    │  DirtyPropagator  │ │  ChangeDetector  │
+    │  (Cascade dirty   │ │  (File-based     │
+    │   state)          │ │   detection)     │
+    └───────────────────┘ └──────────────────┘
+
+Component Responsibilities
+==========================
+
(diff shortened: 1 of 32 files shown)


CybotTM:
Closing this for now. The infrastructure has no consumer: nothing in the repository reads what `ExportsCollectorPass` and `DependencyGraphPass` produce, and both are auto-registered through the `Passes/*Pass.php` glob, so they run on every build without anyone using the result.

It also cannot deliver what it promises in the state it ships in. No production code calls `IncrementalBuildState::setInputDir()`, and with an empty input directory `ExportsCollectorPass` falls back to `hash(serialize($document))` plus `time()`, which `ChangeDetector` then compares against `hashFile()` and the real mtime — two domains that can never match, so every document is dirty on every run and incremental mode silently degrades to no savings at all.

One defect is worth recording separately: `DirtyPropagator::propagate()` builds its dirty set with `array_flip()`, so a document whose name looks numeric becomes an integer key and `array_keys()` then passes an `int` into `getDependents(string $docPath)`. A project with a `2024.rst` — ordinary in changelog and news directories — dies with a `TypeError`; verified against a control document, which passes. Every sibling class guards this with `array_map('strval', array_keys(...))`, this one does not.

I would rather bring this back once there is a consumer for it and those points are addressed, than keep a large unused subsystem open.

PR #1301, opened 2026-02-07
fix: resolve same-page fragment links via `AnchorHyperlinkResolver`

jonathanhefner:
:warning: **Disclaimer:** This was vibe-coded.  I am not a PHP developer, but I am helping https://github.com/modelcontextprotocol/php-sdk set up documentation using phpDocumentor, and I encountered this issue.  (See also https://github.com/modelcontextprotocol/php-sdk/issues/232.)  I guided Claude through the investigation and self-review, but I am unable to personally judge the fix.

---

Same-page fragment links in markdown (e.g., `[Section](#section-one)`) were producing bare `#section-one` URLs because `ExternalReferenceResolver` intercepted all `#`-prefixed references and set bare fragment URLs, preventing `AnchorHyperlinkResolver` from resolving them to canonical URLs.

When combined with a `<base href>` tag in the HTML template (used by phpDocumentor for subdirectory pages), browsers resolve bare fragment URLs against the base URL rather than the current page, causing table-of-contents links to navigate away instead of scrolling.

Remove the `str_starts_with('#')` branch from `ExternalReferenceResolver` so fragment-only references are no longer claimed as "external". Instead, `AnchorHyperlinkResolver` now handles them: it looks up the anchor in the project's internal targets and produces a canonical URL (e.g., `/index.html#section-one`). When no target matches (e.g., RST explicit fragment hyperlinks like `` `link <#abc>`__ ``), it falls back to the bare fragment URL to preserve backward compatibility.

### DIFF
--- a/packages/guides/src/ReferenceResolvers/AnchorHyperlinkResolver.php
+++ b/packages/guides/src/ReferenceResolvers/AnchorHyperlinkResolver.php
@@ -21,11 +21,14 @@ use phpDocumentor\Guides\RenderContext;
 use phpDocumentor\Guides\Renderer\UrlGenerator\UrlGeneratorInterface;
 
 use function count;
+use function str_starts_with;
 
 /**
  * Resolves references with an anchor URL.
  *
- * A link is an anchor if it starts with a hashtag
+ * Looks up the anchor in the project's internal targets and produces a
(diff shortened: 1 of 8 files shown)


jaapio:
Thanks for your help. I decided to pick a different approach to fix this issue. This will be available in our next release.

Issue #1314, opened 2026-03-15
[Bug]: Tilde (`~`) is converted to non-breaking whitespace


wouterj:
### Summary

Not really a 100% bug, but something a bit unexpected as far as I can see.

The `~` character is converted to NBSP by the `InlineLexer`: https://github.com/phpDocumentor/guides/blob/6a008ff1b42f5083e314ce17ae8d4c7129bd92c6/packages/guides-restructured-text/src/RestructuredText/Parser/InlineLexer.php#L134

This behavior was present in the Doctrine RST Parser as well, but I cannot find any reference to this in the Sphinx documentation or reStructured Text specification. A test using Sphinx also confirms that it renders the `~` as tilde character.

I can't find a good way to disable this behavior, given the `InlineLexer` is final and not defined as a service. Overriding the template is also not safe, as `WhitespaceInlineNode` can also be created using the [new `:nbsp:`](https://github.com/phpDocumentor/guides/pull/1280) text role.

Escaping the character in the documentation (`\~`) works fine, but I'm just wondering if this deviation from Sphinx/reStructured Text was intentional :)

### Code snippet that reproduces the problem

```rst
Duration: ~1,000ms

Duration: \~1,000ms
```

### Expected output

```html
<p>Duration: ~1,000ms</p>
<p>Duration: ~1,000ms</p>
```

linawolf:
This was addressed in #1325 (merged), which adds a `disableLegacyTilde` constructor/DI flag to `InlineLexer`/`InlineParser` so the tilde-to-NBSP conversion can be disabled. Note it defaults to `false` (legacy behavior preserved) for backward compatibility, so existing docs aren't silently affected — projects that want the spec-conformant behavior need to opt in via DI config. Closing as resolved.

PR #1324, opened 2026-04-01
Deprecate legacy tilde-to-NBSP behavior

lacatoire:
This addresses #1314.

The tilde character (`~`) was being converted to a non-breaking space (`&nbsp;`), a behavior inherited from `doctrine/rst-parser` that does not match the [reStructuredText specification](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html) or Sphinx behavior.

Following the approach suggested by @jaapio in #1314, this PR adds a `disableLegacyTilde` constructor parameter to `InlineLexer`, passed through `InlineParser` via DI configuration:

- When enabled (default), `~` is treated as a regular `WORD` token
- When disabled, the legacy NBSP behavior is preserved
- The `:nbsp:` text role remains the recommended way to insert non-breaking spaces

### Changes

- `InlineLexer`: added `disableLegacyTilde` constructor parameter, conditional in `getType()`
- `InlineParser`: forwards the flag when instantiating the lexer
- DI config: sets `disableLegacyTilde` to `true` by default
- Unit tests: added tests for both behaviors
- Functional fixtures: updated `nbsp` and `nbsp-role` expected output

### DIFF
--- a/packages/guides-restructured-text/resources/config/guides-restructured-text.php
+++ b/packages/guides-restructured-text/resources/config/guides-restructured-text.php
@@ -373,6 +373,7 @@ return static function (ContainerConfigurator $container): void {
         ->set(DocumentRule::class)
         ->set(InlineParser::class)
         ->arg('$inlineRules', tagged_iterator('phpdoc.guides.parser.rst.inline_rule'))
+        ->arg('$disableLegacyTilde', true)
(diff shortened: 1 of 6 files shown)


jaapio (changes requested):

jaapio (approved):
Thanks a lot for your help.

PR #1327, opened 2026-04-07
Upgrade to phpunit 12

jaapio:


linawolf:
Closing in favor of #1371, which continues this work rebased onto current main (and fixes a PHPUnit 10.5 compatibility bug in the CoversMethod attributes that was failing static analysis here).

PR #1333, opened 2026-04-14
Guard against null token after trailing backslash in text role

### LINKED ISSUE #4082: Warning: Attempt to read property "value" on null
We execute the following command on a documentation project: 
```
docker run --rm -it --pull always \
  -v "./Documentation:/project/Documentation" \
  -v "./Documentation-GENERATED-temp:/project/Documentation-GENERATED-temp" \
  -p 1337:1337 ghcr.io/typo3-documentation/render-guides:latest --config="Documentation" --watch
xdg-open "http://localhost:1337/Index.html"
```
and get the following warning:

Warning: Attempt to read property "value" on null in /opt/guides/vendor/phpdocumentor/guides-restructured-text/src/RestructuredText/Parser/Productions/InlineRules/TextRoleRule.php on line 81

## Expected behavior
No warnings. We expect to have a documentation rendering that is able to update the rendering after saving the file.

## Actual behavior
With warnings. Everything works like expected.

## Steps to reproduce the problem

1. Check out https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6432
2. (ddev composer update) + Execute the command given above

## Your environment

Linux mint cinnamon
[Our code](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6432)

lacatoire:
When a text role candidate is followed by a trailing backslash at the end of the stream (or end of the block), `moveNext()` leaves `$lexer->token` as null and the next read on `->value` triggers a `Warning: Attempt to read property "value" on null`.

Bail out of the loop in that case, matching what the other terminal branches already do. The rule then falls through to `rollback()` and returns null, so the input is simply not interpreted as a text role.

Fixes phpDocumentor/phpDocumentor#4082

### DIFF
--- a/packages/guides-restructured-text/tests/unit/Parser/Productions/InlineRules/TextRoleRuleTest.php
+++ b/packages/guides-restructured-text/tests/unit/Parser/Productions/InlineRules/TextRoleRuleTest.php
@@ -148,4 +148,31 @@ final class TextRoleRuleTest extends TestCase
          */
         self::assertSame($expectedRawContent, $node->rawContent);
     }
+
+    public function testApplyReturnsNullOnTrailingBackslashWithoutRaisingWarning(): void
+    {
+        $input = ":role:`text\\";
+
+        $textRoleFactory = $this->createMock(TextRoleFactory::class);
+        $textRoleFactory->expects(self::never())->method('getTextRole');
+
+        $lexer = new InlineLexer();
+        $lexer->setInput($input);
+        $lexer->moveNext();
+        $lexer->moveNext();
+
+        $textRoleRule = new TextRoleRule();
+        self::assertTrue($textRoleRule->applies($lexer));
+
+        $documentParserContext = new DocumentParserContext(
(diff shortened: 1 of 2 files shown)


Alagts:
Can somebody update this pull request and fix the failing test, please? :)

packages/guides-restructured-text/tests/unit/Parser/Productions/InlineRules/TextRoleRuleTest.php
```diff
@@ -148,4 +148,31 @@ public function __construct(
          */
         self::assertSame($expectedRawContent, $node->rawContent);
     }
+
+    public function testApplyReturnsNullOnTrailingBackslashWithoutRaisingWarning(): void
+    {
+        $input = ":role:`text\\";
```
garvinhicking:
```suggestion
        $input = ':role:`text\\';
```

jaapio:
Thanks for fixing this.

PR #1334, opened 2026-04-16
Deprecate SpecificationInterface in FileCollector and ParseDirectoryCommand

### LINKED ISSUE #1209: Deprecated use of SpecificationInterface on file collector
### Feature request

In version 2 of guides we will remove the support of the `Flyfinder\Specification\SpecificationInterface` in the creation of `phpDocumentor\Guides\Handlers\ParseDirectoryCommand`. We introduced a new `phpDocumentor\FileSystem\Finder\Exclude` that shall be used. This class is used to create the `Flyfinder\Specification\SpecificationInterface` in the FileCollector. 

Users can upgrade using the `Exclude` class from v1.8.0

lacatoire:
refs #1209.

`FileCollector::collect()` now emits a deprecation when a `Flyfinder\Specification\SpecificationInterface` is passed; users should switch to `phpDocumentor\FileSystem\Finder\Exclude` (which the collector already supports). The deprecation message in `ParseDirectoryCommand` is also fixed (concatenation typos, missing spaces) and uses sprintf placeholders.

Side bugfix: `ParseDirectoryHandler` no longer triggers the deprecation when no exclusion is supplied. It used to call `getExcludedSpecification()` unconditionally on the fallback branch, which fired the deprecation even when the caller did nothing wrong. A new `hasExcludedSpecification()` getter on the command lets the handler choose without touching the deprecated path. Internal call sites use `Deprecation::triggerIfCalledFromOutside()` so user-facing notices stay scoped to actual user code.

Tests added for both `FileCollector::collect()` and `ParseDirectoryCommand` covering the three input variants (null, Exclude, SpecificationInterface).

### DIFF
--- a/packages/guides/src/FileCollector.php
+++ b/packages/guides/src/FileCollector.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Guides;
 
+use Doctrine\Deprecations\Deprecation;
 use Flyfinder\Specification\AndSpecification;
 use Flyfinder\Specification\HasExtension;
 use Flyfinder\Specification\InPath;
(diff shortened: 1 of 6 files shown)


jaapio (approved):
Thanks for cleaning up!

PR #1336, opened 2026-04-17
Skip asset copy when destination already has the file

lacatoire:
When an RST project references the same image from multiple documents, `AssetsExtension::copyAsset` calls `$destination->put()` once per reference. On repeat copies the Flysystem Local adapter re-enters `ensureDirectory()` and, on Flysystem v1, logs a noisy `Impossible to create the root directory ... mkdir(): File exist` error even though the HTML output itself is fine (see phpDocumentor/phpDocumentor#3635; v3 self-guards via `is_dir()` before `mkdir`).

Short-circuit the copy when the destination already reports the target path: one extra stat per asset reference, no redundant `put`, no noisy mkdir attempt on repeat renders. The scope of this change is explicit: it silences the error for the *repeat render* case that matches the reporter's symptom. A collision where a non-directory already lives at the parent path (someone else put a bare file there) is out of scope, and the error in that case remains meaningful.

A new `AssetsExtensionTest` covers the skip path, the first-copy path, and idempotency across two consecutive `asset()` calls, and asserts no error is logged on those happy paths.

refs https://github.com/phpDocumentor/phpDocumentor/issues/3635

### DIFF
--- a/packages/guides/src/Twig/AssetsExtension.php
+++ b/packages/guides/src/Twig/AssetsExtension.php
@@ -178,6 +178,11 @@ final class AssetsExtension extends AbstractExtension
                 return $outputPath;
             }
 
+            $destination = $renderContext->getDestination();
+            if ($destination->has($outputPath)) {
+                return $outputPath;
+            }
+
             $fileContents = $renderContext->getOrigin()->read($normalizedSourcePath);
             if ($fileContents === false) {
                 $this->logger->error(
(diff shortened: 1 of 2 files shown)


jaapio:
It is intended that phpDocumentor overwrites existing files. If this triggers warnings we should find a way to prevent it from creating a new directory. I do not think we should keep the original files as this leads to unexpected results for our users.

PR #1345, opened 2026-06-16
[BUGFIX] Read project version from the DOM so "0.10" is not coerced to 0.1

CybotTM:
## Problem

A `<project version="0.10">` in `guides.xml` renders as version `0.1`, and `version="1.0"` as `1`. The digit is lost before anything downstream can object.

## Cause

`XmlUtils::convertDomElementToArray()` runs `phpize()` on every attribute value. That is right for the `<guides>` attributes that want coercion — `links-are-relative`, `max-menu-depth` — and wrong for `<project>`, whose four attributes are all strings.

There was a workaround: write `version="'3.0'"` and let a `beforeNormalization` callback strip the quotes again. It runs *after* `phpize` has already discarded the digit, so it can restore a quoted `3.0` but never an unquoted `0.10`.

## Fix

Read the `<project>` attributes straight from the DOM and detach the element before the conversion, so `phpize` never sees them. The version is correct at the source instead of being coerced and patched up afterwards.

Two consequences that are easy to miss:

- The element is looked up among the **direct** children. The schema lets an `<extension>` carry arbitrary children, and a subtree search would take a `<project>` nested in one for the project configuration and remove it from that extension.
- Detaching `<project>` can leave a root with no children, for which `convertDomElementToArray()` returns `null` rather than an empty array. That is handled rather than asserted, so the outcome does not depend on `zend.assertions`.

The config tree keeps the other half of the removed callback. It preserved the literal of a non-string value, and that half serves every caller, not the XML one: `ContainerFactory::loadExtensionConfig()` feeds a raw PHP array straight in, where a native float reaches a bare `(string)` cast and `(string) 3.0` is `"3"` — the same lost zero, reached without any XML. `null` now passes through instead of becoming the literal string `"NULL"`.

Scope, stated rather than implied: `title` and `copyright` take the same cast on the same path and are not guarded, matching what the callback covered. Only `version` and `release` carry the "always a string" contract and a real-world float shape. The other two writers of a version — `ComposerSettingsLoader` and `VersionFieldListItemRule` — already guard with `is_string`.

## Backward compatibility

Projects that adopted the quoted workaround may still have `<project version="'3.0'">`. To keep those rendering `3.0` and not the literal `'3.0'`, the single quotes are still stripped — now in `XmlFileLoader`, where the quotes come from, and for `version` and `release` only. New files need nothing: write `version="0.10"`.

## Before / after

| Before | After |
| --- | --- |
| before | after |

## Tests

Three integration fixtures, guarding different things. `version-from-guides-xml` uses the unquoted `0.10` it could not express before, and `version-from-guides-xml-nested-project` pins the direct-child lookup; both fail against `main`. `version-from-guides-xml-quoted` cannot fail against `main` — the quoted form worked there, which is its whole point. It guards the other direction and fails as soon as the quote stripping is dropped.

`XmlFileLoaderTest` is new: `guides-cli`'s unit suite covered `Command` and `Logger` only, so the loader was reachable exclusively through fixtures that render a whole site. Eight cases, each seen failing on a defect built for it — DOM reading removed, `trim` widened to every attribute, `trim` dropped, the direct-child search degraded to a subtree search, the project key withheld when empty, and the project key set unconditionally.

The case with no coverage anywhere: **the quote stripping is backward compatibility for two attributes, not a general unquoting rule** — a `title` that really is quoted keeps its quotes.

`GuidesExtensionTest` gains the programmatic-path cases; the float one fails without the fix with `'3'` against `'3.0'`, the null one with `'NULL'` against `''`.

## Verified

Three commits. On the locked set at PHP 8.5, after the rebase onto `main` (33738e22): unit 606 tests / 1216 assertions, functional 120, integration 264, PHPStan and PHPCS clean. Deptrac and the version matrix are left to CI.

## Context

- Reported downstream at TYPO3-Documentation/render-guides#1293

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_014H1xwaAmrQRWUA3vx8bJcD)_

### DIFF
--- a/packages/guides-cli/src/Config/XmlFileLoader.php
+++ b/packages/guides-cli/src/Config/XmlFileLoader.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace phpDocumentor\Guides\Cli\Config;
 
+use DOMAttr;
+use DOMElement;
 use Symfony\Component\Config\Loader\FileLoader;
 use Symfony\Component\Config\Util\Exception\XmlParsingException;
 use Symfony\Component\Config\Util\XmlUtils;
@@ -22,6 +24,7 @@ use function assert;
 use function is_array;
 use function is_string;
 use function sprintf;
+use function trim;
 
 final class XmlFileLoader extends FileLoader
 {
@@ -36,8 +39,44 @@ final class XmlFileLoader extends FileLoader
             throw new XmlParsingException(sprintf('The XML file "%s" is not valid.', $resource));
         }
 
+        // The <project> attributes (title, version, release, copyright) are all
+        // strings and are read from the DOM directly. XmlUtils::convertDomElementToArray()
+        // below runs phpize() on every attribute value, which coerces version-like
+        // strings into numbers ("0.10" would become the float 0.1, "1.0" would
+        // become 1). Reading them straight from the DOM and detaching <project>
+        // beforehand keeps the version exactly as written.
+        $projectConfig = null;
+        $project = $element->getElementsByTagName('project')->item(0);
+        if ($project instanceof DOMElement) {
+            $projectConfig = [];
+            foreach ($project->attributes as $attribute) {
+                if (!($attribute instanceof DOMAttr)) {
+                    continue;
+                }
+
+                $value = $attribute->value;
+
+                // Backward compatibility: to stop the previous phpize() call from
+                // turning a version into a number, consumers wrapped it in single
+                // quotes (e.g. version="'3.0'"). The value is now read straight from
+                // the DOM so the quotes are no longer needed, but existing guides.xml
+                // files may still contain them; strip them for these two attributes.
+                if ($attribute->name === 'version' || $attribute->name === 'release') {
+                    $value = trim($value, "'");
+                }
+
+                $projectConfig[$attribute->name] = $value;
+            }
+
+            $project->parentNode?->removeChild($project);
+        }
+
         $rootConfig = XmlUtils::convertDomElementToArray($element);
         assert(is_array($rootConfig));
+
+        if ($projectConfig !== null) {
+            $rootConfig['project'] = $projectConfig;
+        }
 
         $configs = [];
         if (isset($rootConfig['import'])) {
--- a/packages/guides/src/DependencyInjection/GuidesExtension.php
+++ b/packages/guides/src/DependencyInjection/GuidesExtension.php
@@ -43,11 +43,8 @@ use function array_values;
 use function assert;
 use function dirname;
 use function is_array;
-use function is_int;
 use function is_string;
 use function pathinfo;
-use function trim;
-use function var_export;
 
 final class GuidesExtension extends Extension implements CompilerPassInterface, ConfigurationInterface, PrependExtensionInterface
 {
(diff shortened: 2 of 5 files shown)


tests/Integration/tests/meta/version-from-guides-xml/input/guides.xml
```diff
@@ -6,7 +6,7 @@
 >
     <project
             title="Render guides"
-            version="'3.0'"
+            version="0.10"
```
garvinhicking, line 9: I think to cover for regressions and others we also need to have a file in place that uses the pre-quote (since consumers might have put this in already)

CybotTM: as ou please

CybotTM: Good call — added `tests/Integration/tests/meta/version-from-guides-xml-quoted` with `version="'3.0'"` and `release="'3.0.0'"`. To keep guides.xml files that adopted that workaround rendering `3.0` (not the literal `'3.0'`), the surrounding single quotes are still stripped for `version`/`release` — the same `trim($value, "'")` the removed `beforeNormalization` did, now in `XmlFileLoader`. Both the quoted and the direct `0.10` forms are green.

Proof: quoted version="'3.0'" renders version 3.0

garvinhicking:
Thanks - I can't really judge the impact of this and hope @jaapio can give some feedback. I remember having stabbed at this and not being able to resolve this.

linawolf:
This PR fixes trailing-zero loss (`3.0` -> `"3"`) by moving `version`/`release` handling out of XML normalization into `XmlFileLoader`, but the `beforeNormalization` callback removed here wasn't XML-specific -- it also caught non-string scalars reaching this config tree from anywhere else.

`ContainerFactory::loadExtensionConfig()` feeds a raw PHP array straight into this same tree, bypassing `XmlFileLoader` entirely. Before this PR, `beforeNormalization`'s `!is_int && !is_string` branch caught that case; after, nothing does, so `['project' => ['version' => 3.0]]` -- a native PHP float -- hits a bare `(string)` cast downstream and loses its trailing zero again (`(string) 3.0` -> `"3"`), just via the programmatic path instead of XML.

`GuidesExtensionTest.php` already has the right harness for this (it calls `$extension->load($configs, $container)` with raw arrays) -- just missing a case for it:
```php
yield 'project version as float' => [
    [['project' => ['version' => 3.0]]],
    static function (ContainerBuilder $container): void {
        self::assertSame('3.0', $container->getParameter('phpdoc.guides.project.version'));
    },
];
```

CybotTM:
After some back-and-forth, I haven't managed to significantly reduce the scope of this PR. I’ve tried to structure the PR logically based on the commits, hopefully, that makes the review easier.

jaapio:
Small note, I noticed that the referenced issue is pointing to something different. Which makes it harder to see if this was an issue. Given the fact code changes are made I assume there was an issue. But I could not find it. If you have the correct issue I can read up from the beginning. 

I will continue to review your changes step by step.

CybotTM:
Hi @jaapio ,

thanks for taking care of.

> Small note, I noticed that the referenced issue is pointing to something different.

my fault, I mixed something while working on different issues.

Correct: TYPO3-Documentation/render-guides#1293

you can see the issue here for example: https://docs.typo3.org/p/netresearch/nr-llm/0.10/en-us/ 

image

jaapio (approved):
This seems to be a solid fix. Thank for that. Could you please resolve the conflicts and do a rebase so we can merge this?

CybotTM:
Rebased onto `main` (33738e22). The baseline commit is gone: #1386's PHPUnit 12 upgrade (0f28da81) regenerated `phpunit-baseline.xml` without the vendor entries, so the `vendor/symfony/config/Loader/FileLoader.php` block it dropped no longer exists. Three commits now; the description section about it is removed.

jaapio:
Thanks @CybotTM!

PR #1351, opened 2026-08-15
[FEATURE] Support marker based partial include in literalinclude

CybotTM:
## Problem

`literalinclude` currently always includes the complete referenced file. An author who wants to show one relevant region of a real source or configuration file therefore has to duplicate that region into a dedicated snippet file, and that copy then silently drifts away from the original, or include unrelated content around the interesting part.

This came out of a TYPO3 documentation pull request ([TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument#547](https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/pull/547)) that documented the Sphinx options for partial inclusion. It was closed as wrong, because the documented behaviour does not exist here — the missing piece is the feature, not the documentation.

## What this adds

The Sphinx options `:start-after:` and `:end-before:`:

```rst
..  literalinclude:: Example.php
    :language: php
    :start-after: // begin example
    :end-before: // end example
```

The included region starts on the line following the first line containing the `start-after` text and ends on the line preceding the first line containing the `end-before` text. `end-before` is searched behind the start of the region, so the same marker text may be used more than once in one file.

Markers were chosen over line numbers because they stay valid when the file changes above or below the selected region.

## Behaviour when a marker is not found

A warning is logged (`rst-file` and line number included, as usual) and nothing is included. Falling back to the complete file would publish exactly the content the option was meant to exclude, and a build using `--fail-on-log` still fails on the warning. The same applies when an option is used without a value, when `end-before` occurs only above the line matched by `start-after` — that case gets its own message, because reporting the marker as "not found" for a marker that is plainly in the file sends the reader hunting for a typo that is not there — and when the marked region turns out to be empty, which is what a stale marker left directly under its partner looks like.

## Notes for review

* The selection happens before the `CodeNode` is created, so it lives in `LiteralincludeDirective` and not in `DefaultCodeNodeOptionMapper`. That mapper is shared with `code-block`, and it maps display options (language, caption, emphasized lines, line numbering) rather than changing the input.
* `LiteralincludeDirective` now takes a logger. It is optional and defaults to null, following `ConfvalDirective`, which is the one place in this directory where a logger was added to an existing directive; that keeps the constructor change additive for anyone building the directive outside the container. No change to the service definition was needed, the container autowires it.
* Six integration tests cover: `start-after` alone, `end-before` alone, both together, a marker that does not exist, markers given in the wrong order, and an empty marked region.
* The second commit adds `docs/reference/restructuredtext/literalinclude.rst`. The reference section had no page for `literalinclude` at all, so the new options had nowhere to be documented; the page therefore also covers the directive itself and the options the shared code node option mapper provides. Every documented behaviour was rendered with `vendor/bin/guides` and read back before it was written down, including the missing file case (error logged, nothing rendered) and the missing marker case (warning logged, nothing included).
* `:diff:` is deliberately out of scope, it needs a different data and render model.
* The Sphinx option `:lines:` is proposed separately in [#1352](https://github.com/phpDocumentor/guides/pull/1352), which is stacked on this branch.

Local run of `phpunit` (unit, functional, integration), `phpcs`, `phpstan` and `deptrac`: green.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_015QXXkquh2eQNBiTYA39Wss)_

### DIFF
--- a/docs/reference/restructuredtext/literalinclude.rst
+++ b/docs/reference/restructuredtext/literalinclude.rst
@@ -1,0 +1,75 @@
+..  include:: /include.rst.txt
+
+..  _literalinclude:
+
+==============
+Literalinclude
+==============
+
+The ``literalinclude`` directive includes the content of another file as a code block. It lets the documentation
+show a real, working file instead of a copy that has to be kept in sync by hand.
+
+..  code-block::
+
+    ..  literalinclude:: Example.php
+        :language: php
+
+The path is resolved relative to the document that contains the directive, a path starting with ``/`` relative to
+the root of the documentation. If the file cannot be read, an error is logged and nothing is rendered in its place.
+Rendering with ``--fail-on-error`` turns that error into a failing build.
+
+Including only a part of a file
+===============================
+
+Often only one region of a file is worth showing. The options ``:start-after:`` and ``:end-before:`` select that
+region by the text of the lines enclosing it:
+
+..  code-block::
+
+    ..  literalinclude:: Example.php
+        :language: php
+        :start-after: // begin example
+        :end-before: // end example
+
+The region starts on the line *following* the first line that contains the ``start-after`` text, and ends on the
+line *preceding* the first line that contains the ``end-before`` text. Both marker lines are left out. The
+``end-before`` text is searched behind the start of the region, so the same marker may occur several times in one
+file.
+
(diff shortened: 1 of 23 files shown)


jaapio (approved):

jaapio:
Thanks for this addition!

PR #1352, opened 2026-08-15
[FEATURE] Support :lines: in literalinclude

CybotTM:
Rebased onto `main` now that [#1351](https://github.com/phpDocumentor/guides/pull/1351) has been merged, so the two commits here are the only ones left. @linawolf's two review points are addressed and the change has had another pass since; ready for review.

## What this adds

The Sphinx option `:lines:` for `literalinclude`, so a region can also be selected by line number:

```rst
..  literalinclude:: Example.php
    :language: php
    :lines: 1,3-5,20-
```

Line numbers are 1 based, ranges are inclusive, an omitted end means "up to the last line", and overlapping ranges are included once, in the order of the file. Both ends are clamped to the lines that exist, so `:lines: 1-99999999999` selects the whole region instead of iterating over the number the author wrote down.

The accepted spelling is validated with `DefaultCodeNodeOptionMapper::LINE_NUMBER_RANGES_REGEX`, the pattern already used for `:emphasize-lines:`, so both options are written the same way. If you would rather not have the directive reference the mapper's constant, say so and I will move the pattern to a shared place.

## Order of application

Like Sphinx, `:lines:` is applied **after** `:start-after:` and `:end-before:`, so the numbers count within the marked region rather than within the file. There is an integration test for that combination.

One deliberate deviation from Sphinx: Sphinx keeps the order the ranges were written in and repeats a line that two ranges both select, so `:lines: 5,1` yields line 5 followed by line 1. This implementation sorts and de-duplicates, because a code sample read out of order is far more likely to be a typo than an intention. It is documented on the reference page and pinned by an integration test.

I would still describe markers as the more valuable half of this: line numbers break on every edit above the selected region, markers do not. `:lines:` is here for Sphinx compatibility and for files that have no natural marker.

## Behaviour on an invalid or empty selection

Each range that selects no line is warned about on its own, naming that range, and the remaining ranges are still included — `:lines: 1-2,100` on a 14 line file includes lines 1 and 2 and says that `100` selected nothing. A missing value or a value that is not a list of line numbers logs a warning and includes nothing, consistent with a marker that is not found in [#1351](https://github.com/phpDocumentor/guides/pull/1351).

## Interaction with the numbering options

`:emphasize-lines:` counts within what `:lines:` left over, and `:lineno-start:` stays a plain override of the first number displayed, unrelated to where the selected lines sat in the file. Neither was stated on the reference page and neither was pinned by a test; both are now.

## Documentation

The reference page added in [#1351](https://github.com/phpDocumentor/guides/pull/1351) gets a section on selecting lines by number, including the order of application and the advice to prefer markers where a file has them.

## Tests

Eleven integration tests: a mixed range list, an open ended range, an invalid specification, a range beyond the end of the file, a range list where only one entry is beyond the end, `:lines:` used without a value, a range whose end exceeds `PHP_INT_MAX` as a regression test for the clamping, overlapping ranges given out of order, `:lines:` combined with markers, `:lines:` combined with markers and `:lineno-start:`, and `:lines:` combined with `:emphasize-lines:`.

Local run on the rebased tree: `phpunit` (935 tests, unit + functional + integration), `phpcs`, `phpstan` and `deptrac` all green, `tools/xmllint.sh` clean.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Mf63edGvCVRQz6mwF8gxcC)_

### DIFF
--- a/docs/reference/restructuredtext/literalinclude.rst
+++ b/docs/reference/restructuredtext/literalinclude.rst
@@ -64,11 +64,15 @@ are included once, in the order of the file.
 ``:lines:`` is applied *after* ``:start-after:`` and ``:end-before:``, so when they are combined the numbers count
 within the marked region rather than within the whole file.
 
+``:emphasize-lines:`` in turn counts within what ``:lines:`` left over, and ``:lineno-start:`` remains a plain
+override of the first number displayed, unrelated to where the selected lines sat in the file.
+
 Prefer markers where a file has them: line numbers change with every edit above the selected region, and nothing in
 the source file records that the documentation depends on them.
 
-If the value cannot be read as a list of line numbers, or if it selects no line at all, a warning is logged and
-nothing is included.
+Every range is warned about separately when it selects no line, for example because it lies beyond the end of the
+region; the remaining ranges are still included. If the option is used without a value, or its value cannot be read
+as a list of line numbers, a warning is logged and nothing is included.
 
 Options
 =======
@@ -90,7 +94,8 @@ Options
     and ``end-before``.
 
 ``:emphasize-lines:``
-    Line numbers to be highlighted, for example ``3,5-6``. Counted within the included region.
+    Line numbers to be highlighted, for example ``3,5-6``. Counted within the included region, that is within what
+    is left after ``start-after``, ``end-before`` and ``lines``.
 
 ``:linenos:``
     Displays line numbers, starting at 1.
--- a/docs/reference/restructuredtext/literalinclude.rst
+++ b/docs/reference/restructuredtext/literalinclude.rst
@@ -47,6 +47,29 @@ publish the very content it was meant to exclude. The same happens when the mark
 when the ``end-before`` text occurs only above the line matched by ``start-after``. Rendering with ``--fail-on-log``
 turns any of these warnings into a failing build.
 
+Selecting lines by number
+=========================
+
+A region can also be selected by line number, which is useful for files that have no natural marker:
+
+..  code-block::
+
+    ..  literalinclude:: Example.php
+        :language: php
+        :lines: 1,3-5,20-
+
+Line numbers start at 1, ranges are inclusive, and an omitted end means "down to the last line". Overlapping ranges
+are included once, in the order of the file.
+
+``:lines:`` is applied *after* ``:start-after:`` and ``:end-before:``, so when they are combined the numbers count
+within the marked region rather than within the whole file.
+
+Prefer markers where a file has them: line numbers change with every edit above the selected region, and nothing in
+the source file records that the documentation depends on them.
+
+If the value cannot be read as a list of line numbers, or if it selects no line at all, a warning is logged and
+nothing is included.
+
 Options
 =======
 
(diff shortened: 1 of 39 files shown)


linawolf:
Coverage gap: there are existing tests for `:linenos:`/`:lineno-start:` in isolation, but none combine it with the new `:lines:`/`:start-after:`/`:end-before:` filtering. Checked against Sphinx's `LiteralIncludeReader` -- `:lineno-start:` is meant to be a plain override independent of filtering (no `:lineno-match:` equivalent exists here), so I believe current behavior is already correct, just untested. Worth adding one fixture combining them, e.g. `:start-after:` + `:lines:` + `:lineno-start: 10`, asserting the output numbers from 10 regardless of where the selected content sat in the source file.

linawolf:
`literalinclude-lines-huge-range` uses `:lines: 1-99999999999` (~10^11) as the regression test for clamping "far beyond PHP_INT_MAX territory," but that value doesn't actually exceed PHP_INT_MAX on 64-bit systems (~9.2x10^18) -- it's off by about 7 orders of magnitude. A genuinely overflowing value clamps correctly today (checked manually), but that path has no test coverage, so a future change to `parseRange()`'s arithmetic could silently reintroduce overflow behavior undetected. Suggest bumping the value to something that actually exceeds PHP_INT_MAX on 64-bit systems, e.g. `1-99999999999999999999`.

jaapio:
@CybotTM could you please rebase this branch against main? So we have a clean review

CybotTM:
Both confirmed and fixed in 6323f099.

On the coverage gap: right on both counts. There was no fixture combining the numbering options with any filtering, and your reading of the intent matches what the code does — `DefaultCodeNodeOptionMapper::setStartingLineNumberBasedOnOptions()` only ever writes `data-start` on the node and never looks at the content, so `:lineno-start:` is a plain override with no `:lineno-match:` equivalent. Added `literalinclude-lines-lineno-start`, exactly the combination you proposed: `:start-after:` + `:end-before:` + `:lines: 1-4` + `:linenos:` + `:lineno-start: 10`, asserting `data-start="10"` on content that sits at lines 8 to 11 of the source file. I injected a defect into it (`data-start="1"`) and confirmed it fails there, so it pins the thing it claims to pin rather than merely rendering.

On the overflow value: you are right and the number was wrong by the margin you name. `99999999999` is about 10^11 against a 64-bit `PHP_INT_MAX` of about 9.2 * 10^18. I measured the path you checked by hand: `(int) "99999999999999999999"` saturates to `9223372036854775807`, which `min()` then clamps to the line count, so the behaviour is correct today and was, as you say, untested. Raised the value in `literalinclude-lines-huge-range` to `1-99999999999999999999` rather than adding a second fixture, since the smaller value's clamping is already covered by `literalinclude-lines-out-of-range` and `literalinclude-lines-partly-out-of-range`; the fixture now earns its name. The claim in the pull request description that the old value was "far beyond `PHP_INT_MAX` territory" was simply false and I have corrected it.

Also added `literalinclude-lines-overlapping` while re-reading the change, because the reference page asserts that overlapping ranges are included once in the order of the file and nothing pinned that either.

CybotTM:
Rebased onto current `main` — [#1351](https://github.com/phpDocumentor/guides/pull/1351) had merged in the meantime, so its two commits dropped out and the two commits left here are the whole change. New head is 6323f099. Also picked up @linawolf's two review points in the same push.

jaapio (approved):

jaapio:
Thanks for this addition, code looks good to me, I did a rebase and enabled the automerge. Will be merged soon.

Issue #1355, opened 2026-08-19
Split packages do not declare the dependencies their code uses


CybotTM:
### Summary

Every split package declares fewer dependencies than its production code uses. The classes resolve inside the monorepo because the root pulls everything in together, but each `packages/*` directory is published as a standalone Composer package, and there the metadata is simply wrong. A consumer installing `phpdocumentor/filesystem` on its own gets a `FlySystemAdapter` whose `League\Flysystem` and `Flyfinder` classes are not required by anything.

I ran `composer-require-checker` against each split package in isolation — package directory copied out, its own `composer install`, `autoload-dev` and `require-dev` removed so only production code is scanned. Unknown symbols per package:

| Package | unknown symbols |
| --- | --- |
| phpdocumentor/guides | 26 |
| phpdocumentor/guides-restructured-text | 28 |
| phpdocumentor/filesystem | 16 |
| phpdocumentor/guides-cli | 16 |
| phpdocumentor/guides-graphs | 12 |
| phpdocumentor/guides-code | 11 |
| phpdocumentor/guides-theme-rst | 11 |
| phpdocumentor/guides-markdown | 10 |
| phpdocumentor/dev-server | 8 |
| phpdocumentor/guides-theme-bootstrap | 6 |

`psr/log` is already handled separately in #1354 and no longer appears in these numbers.

### Straightforward part

Most of it is a plain missing `require` for a library the code uses directly and that is already installed transitively. Nothing optional, no semantics involved:

| Package | missing |
| --- | --- |
| filesystem | `league/flysystem`, `phpdocumentor/flyfinder` |
| guides | `psr/container`, `psr/clock`, `league/uri-interfaces`, `symfony/http-client-contracts` |
| guides-cli | `psr/container`, `psr/clock`, `psr/event-dispatcher`, `doctrine/deprecations`, `league/tactician`, `phpdocumentor/flyfinder`, `phpdocumentor/filesystem`, `webmozart/assert` |
| guides-code | `twig/twig` |
| guides-graphs | `symfony/http-client-contracts`, `webmozart/assert` |
| guides-markdown | `symfony/string` |
| guides-restructured-text | `symfony/string`, `phpdocumentor/flyfinder` |
| guides-theme-rst | `twig/twig`, `symfony/polyfill-mbstring` |
[...]

PR #1356, opened 2026-08-19
[TASK] Declare the direct dependencies each package uses

CybotTM:
Implements the mechanical half of #1355; everything that needs a decision stays there.

## Problem

Every split package uses libraries in its production code without requiring them; this changes the nine where the missing entry needs no decision. The classes resolve inside the monorepo because the root pulls everything in together, but every `packages/*` directory is published as its own Composer package, and there the metadata is wrong. `phpdocumentor/filesystem` is the clearest case: it ships a `FlySystemAdapter` built on `League\Flysystem` and `Flyfinder`, and requires neither.

## What this adds

27 declarations across nine packages. Every library added here is already installed transitively today, so no consumer loses a resolution, and the constraints follow what this repository declares elsewhere. Existing entries keep their order; nothing is re-sorted.

| Package | added |
| --- | --- |
| filesystem | `league/flysystem`, `phpdocumentor/flyfinder` |
| guides | `league/uri-interfaces`, `psr/clock`, `psr/container`, `symfony/http-client-contracts` |
| guides-cli | `doctrine/deprecations`, `league/tactician`, `phpdocumentor/filesystem`, `phpdocumentor/flyfinder`, `psr/clock`, `psr/container`, `psr/event-dispatcher`, `webmozart/assert` |
| guides-code | `twig/twig` |
| guides-graphs | `symfony/http-client-contracts`, `webmozart/assert` |
| guides-markdown | `symfony/string` |
| guides-restructured-text | `phpdocumentor/flyfinder`, `symfony/string` |
| guides-theme-rst | `symfony/polyfill-mbstring`, `twig/twig` |
| dev-server | `guzzlehttp/psr7`, `psr/event-dispatcher`, `ratchet/rfc6455`, `react/event-loop`, `symfony/event-dispatcher-contracts` |

The one entry that is not simply a library the code imports: `guides-theme-rst` calls `mb_str_pad()` in `RstTheme/Twig/RstExtension.php:154`, which PHP provides only from 8.3 while the package allows `^8.1`. `symfony/polyfill-mbstring` supplies it from v1.28 and already arrives through `symfony/string` and `twig/twig`; this declares it. Verified by installing the package standalone in a PHP 8.1 container.

`psr/log` was #1354, which has since merged; this branch is rebased on top of it and both columns above are measured against that `main`.

## Verification

Measured, not assumed. Every package was copied out, installed standalone with `autoload-dev` and `require-dev` removed, and checked with `composer-require-checker` — once against `main` and once against this branch, so both columns share one baseline:

| Package | on `main` | on this branch |
| --- | --- | --- |
| dev-server | 8 | 0 |
| guides-cli | 16 | 4 |
| guides-theme-rst | 11 | 6 |
| filesystem | 16 | 6 |
| guides | 26 | 21 |
| guides-graphs | 12 | 9 |
| guides-restructured-text | 28 | 25 |
| guides-code | 11 | 9 |
| guides-markdown | 10 | 9 |
| guides-theme-bootstrap | 6 | 6 |

`guides-theme-bootstrap` is unchanged because its only gap is the Symfony config and container pair, which stays in #1355. Every symbol still listed elsewhere belongs to a decision item there, to `psr/log` from #1354, or to docblock pseudo-types that belong in `composer-require-checker.json`.

Note for anyone re-running this: the `composer-require-checker` version pinned in `.phive/phars.xml`, 3.5.1, aborts on `symfony/config` v8 with a parse error and produces empty output that reads like a pass. These numbers come from `composer-require-checker` 4.24.0.

Full suite 828 tests, no failures, also with `--prefer-lowest`. PHPStan and PHPCS clean. `composer validate --strict` passes for the root and every package.


linawolf:
Small non-blocking note on `react/event-loop: ^1.5` in `dev-server/composer.json`: the symbols this package actually uses — `React\EventLoop\Loop` (`Server.php`) and `LoopInterface` (`INotifyWatcher.php`) — have both been available since **1.2.0**, per the package's own changelog:

> ## 1.2.0 (2021-07-11)
> Feature: Introduce new concept of default loop with the new `Loop` class.

`composer-require-checker` confirms a symbol exists in whatever's currently installed — it doesn't derive the true minimum version, so a floor pinned there just reflects today's lockfile rather than the actual requirement. Worth either loosening this one to `^1.2` or leaving a short note that the floor is "checker-verified, not independently confirmed minimum" for anyone tightening these further later, so the same shortcut doesn't quietly become an assumed guarantee.

jaapio:
Thanks for this cleanup, I think it would be good if we also implement a workflow in github to check the packages, because now they are updated, but they will be running behind at some point again. Especially because we also run dependabot with updates.

CybotTM:
Hey @jaapio ,

> I think it would be good if we also implement a workflow in github to check the packages

maybe helpful: https://github.com/maglnet/ComposerRequireChecker

PR #1357, opened 2026-08-19
[TASK] Move TestExtension out of the shipped source

CybotTM:
Moves `TestExtension` from `packages/guides/src/` to `packages/guides/tests/unit/`. Pure rename, no content change.

## Why

It is test infrastructure, not library code. `process()` forces `Parser`, `Compiler` and `phpdoc.guides.output_node_renderer` public so tests can pull them from the container, replaces the clock with `MockClock` pinned to `2023-01-01 12:00:00`, and registers Monolog's `TestHandler` so functional tests can assert on log records. Its only user is `tests/ApplicationTestCase.php`, from which `tests/Functional/FunctionalTest.php` and `tests/Integration/IntegrationTest.php` inherit. No `packages/*/src` references it.

It reached this package by accident rather than by decision. `1d10cbd8` created it in `packages/guides-symfony/src/DependencyInjection/` next to the new `ContainerFactory`, wiring it into `ApplicationTestCase` in the same commit. `54daa48f` dissolved `guides-symfony` the next day, and the rename carried the file into `packages/guides`. `ba041eb8` then added the Monolog handler and `541fbf4d` the mocked clock, each without revisiting where the class lives, and the "make all classes final" sweep in `a7395d07` treated it as ordinary library code.

The effect is that `phpdocumentor/guides` ships a class using `Monolog\Logger`, `Monolog\Handler\TestHandler` and `Symfony\Component\Clock\MockClock` as production symbols, none of which the package requires. That is one of the items in #1355.

## Why this is safe

The fully qualified name is unchanged. `packages/guides/tests/unit/` maps to the same `phpDocumentor\Guides\` namespace as `src/`, and the root `autoload-dev` already lists that path, so no import, no consumer and no configuration changes. Because every package marks `/tests export-ignore`, the class simply stops being part of the published `phpdocumentor/guides`.

The one visible change for a consumer: anyone who pulled `phpDocumentor\Guides\DependencyInjection\TestExtension` out of the published package to test their own extension would lose it. It is undocumented and appears in no guide, so this looks unintended rather than a supported entry point — worth a second opinion if you disagree.

## Verification

Full suite 828 tests, no failures. PHPStan and PHPCS clean.


wouterj:
This is a BC break. I'm using the test extension in the test suite for the Symfony docs builder, and it can no longer use it.

I don't see any problem with shipping test helpers in a package, to make testing easier for users. Besides that, there is no dependency here. `::class` always resolves, it doesn't need the class to exist. At most, the container will have a service definition that is not used - which will automatically get removed by the compiler pass. Imho, this is just a busywork cleanup with no real benefit :)

CybotTM:
Hi @wouterj,

Thanks for pointing out the downstream usage. I treated this as infrastructure for the packages own tests, but the lack of other references within this repository was not enough to establish that.

You’re right about ::class: those references do not require the classes to exist just to resolve their names. The helper still needs the corresponding services when activated, but that is different from imposing those requirements on every consumer.

Given that the class is useful to downstream tests, I suggest reverting the move and documenting it as a testing utility, including its prerequisites. Separating the packages own tests from shipped code should not mean removing useful test support from consumers.

CybotTM:
@wouterj I provided #1397 - you may comment there your support for it if you like. It also adds documentation around it.

PR #1358, opened 2026-08-23
[FEATURE] Support index generation (genindex)

### LINKED ISSUE #921: Support Index generation (Glossary)
### Feature request

Sphinx allows to create indexes like this:

https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/genindex.html

This interprets RST tagged indices:

```
.. index:: 
   XLIFF; Files
   File; XLIFF
```

and with that information creates a reference index as "genindex.html".

Having an index like this helps readers of documentation to easily see which terms have been referenced on which page, where manual indexes have been placed.

linawolf:
Sphinx-authored documentation -- including TYPO3 Core's own Changelog
files -- relies on `.. index::` entries so readers can jump straight to
every page a term is discussed on, the way a printed book's index does.
This library parsed those entries but threw them away, so porting such
docs here meant silently losing that navigation aid.

Entries are now collected project-wide and made available as a genindex
page, so documentation migrated from Sphinx keeps working the way its
authors intended.

Each term is also stamped onto the section it belongs to as a data
attribute, independent of any genindex page. That's for tooling that
crawls the rendered HTML rather than the reST source -- a custom search
engine such as TYPO3's Elasticsearch integration can pick up the same
terms an author already tagged with `index`, instead of reimplementing
that logic itself.

Closes #921

Assisted-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Y48g2S9Q2uUVLcsTEAeyGX
Signed-off-by: linawolf


packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
```diff
@@ -13,10 +13,146 @@
...
+
+        $segments = [];
+        foreach ($lines as $line) {
+            foreach (explode(',', $line) as $segment) {
+                $segments[] = $segment;
+            }
+        }
+
+        $segments = array_values(array_filter(array_map(trim(...), $segments), static fn (string $segment): bool => $segment !== ''));
+
```
jaapio: I think this line trims all segments and removes the empty ones. But that's not really clear from this line. The code seem to be a duplicate from what happens on line 86. Maybe it's better to wrap this in a method and reuse that?

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
```diff
@@ -13,10 +13,146 @@
...
+use function sprintf;
+use function str_starts_with;
+use function strtolower;
+use function substr;
+use function trim;
+
+/**
+ * Collects index entries, example:
+ *
+ * .. index:: single: installation
```
jaapio, line 39: is this a valid way of writing an directive?

packages/guides/src/Nodes/Metadata/TemplateNode.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @link https://phpdoc.org
+ */
+
+namespace phpDocumentor\Guides\Nodes\Metadata;
+
+/**
+ * Selects an alternate template for rendering the document, e.g. `:template: genindex`.
+ */
+final class TemplateNode extends MetadataNode
```
jaapio, line 19: I think this could be a separate PR, which introduced the logic to define a different template for a document. That would make this PR impact smaller, and easier to review.

packages/guides/src/Compiler/Passes/IndexCollectorPass.php
```diff
@@ -0,0 +1,600 @@
...
+use function count;
+use function implode;
+use function mb_strtolower;
+use function sprintf;
+use function str_starts_with;
+use function strcasecmp;
+use function trim;
+use function usort;
+
+/**
```
jaapio, line 27: it feels like this class is doing to much, it does collect the entries, resolve them and add them to the genIndexNode. That seems to be lot. 

Maybe we can separate the functionalities. That would also allow other compilerpasses to add entries to the genindex or remove them when filtering is required.

packages/guides-restructured-text/src/RestructuredText/Directives/GenIndexDirective.php
```diff
@@ -0,0 +1,69 @@
...
+use function trim;
+
+/**
+ * Renders a genindex term listing wherever it's placed, optionally scoped to
+ * documents under one or more path prefixes -- unlike the project-wide
+ * `:template: genindex` page, this can be dropped inline anywhere and used
+ * more than once, e.g. one per version directory in a changelog:
+ *
+ * .. genindex::
+ *
```
jaapio: if we add examples in a code block we can show them in the docs once we get there.

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
```diff
@@ -13,10 +13,146 @@
...
+ *
+ * .. index::
+ *      single: configuration
+ *      pair: configuration; file
+ *      see: token; access token
+ *      ! main entry example
+ *
+ * A line may also hold several comma-separated, type-less entries at once
+ * (the convention used by e.g. TYPO3's Core Changelog files), each becoming
+ * its own "single" entry:
```
jaapio: is this specific for TYPO3? or is the comment here just to narrow?

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
```diff
@@ -13,10 +13,146 @@
...
+                continue;
+            }
+
+            $closestType = $case;
+            $closestDistance = $distance;
+        }
+
+        if ($closestType === null || $closestDistance > self::TYPO_DISTANCE_THRESHOLD) {
+            return;
+        }
```
jaapio:
```suggestion
        if ($closestType === null) {
            return;
        }
        
        if ( $closestDistance > self::TYPO_DISTANCE_THRESHOLD) {
            return;
        }
```

packages/guides/src/Nodes/Index/GenIndexRow.php
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * One rendered row of a genindex term: either a direct link to a section
+ * ("link"), or a "see"/"seealso" cross-reference to another term.
+ *
+ * @extends AbstractNode<ReferenceNode|null>
+ */
+final class GenIndexRow extends AbstractNode
+{
+    public function __construct(
+        private readonly GenIndexRowKind $kind,
```
jaapio, line 28: If this is needed, why not create a type per `GenIndexRow` type?  `GenIndexLink`, `GenIndexSee`  then you can create a template per type, not having a template with switches in it.

packages/guides/resources/template/html/body/genindex/row.html.twig
```diff
@@ -0,0 +1,10 @@
+{% if node.isLink %}
```
jaapio: we could switch here on `getKind` no need to have this separate methods.

packages/guides/src/NodeRenderers/Html/GenIndexTermNodeRenderer.php
```diff
@@ -0,0 +1,44 @@
...
+use phpDocumentor\Guides\NodeRenderers\NodeRenderer;
+use phpDocumentor\Guides\Nodes\Index\GenIndexTerm;
+use phpDocumentor\Guides\Nodes\Node;
+use phpDocumentor\Guides\RenderContext;
+use phpDocumentor\Guides\TemplateRenderer;
+
+use function is_a;
+
+/** @implements NodeRenderer<GenIndexTerm> */
+final class GenIndexTermNodeRenderer implements NodeRenderer
```
jaapio: This is just a normal templateRenderer right? So adding the template to the template map is sufficient.

jaapio (changes requested):
I think this is a great  addition to what we already do with this project. However I would love to see a bit smaller PR in steps. I think the template on documents makes a lot of sense also in a separate feature. 

The compiler pass needs some extra attention. It contains a lot of logic, which could be split. This would allow tools like phpDocumentor to extend the index before rendering.

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
linawolf: done

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
linawolf: Yes, valid — it matches Sphinx's own single-line directive syntax. Sphinx's docs source (`doc/usage/restructuredtext/basics.rst`) uses the exact same pattern:

```rst
.. index:: ! pair: global; substitutions
```

i.e. `.. index:: <type>: <entry>` on one line, same as `.. index:: single: installation` here. `process()` already handles this form via `$directive->getData()` (falling back to the indented block form when the directive line itself is empty). TYPO3's own changelog files use this pattern too.

Source: https://github.com/sphinx-doc/sphinx/blob/master/doc/usage/restructuredtext/basics.rst#L578

packages/guides/src/Nodes/Metadata/TemplateNode.php
linawolf: Done — split into #1364. That branch now owns `TemplateNode`, `TemplateFieldListItemRule`, `TemplateMetadataNodeRenderer`, and `DocumentNode::getTemplate()`; this PR is rebased on top of it, so its diff now covers genindex-specific work only. Once #1364 merges into `main`, this PR's base will follow automatically.

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
linawolf: rephrased it

packages/guides/src/NodeRenderers/Html/GenIndexTermNodeRenderer.php
linawolf: done

packages/guides/src/Compiler/Passes/IndexCollectorPass.php
linawolf: Split into four collaborators under a new `Compiler/Passes/IndexCollector/` namespace: `IndexEntryCollector` (collect + expand `.. index::` entries into a term map), `GenIndexSeeResolver` (resolve `see`/`seealso` targets), `GenIndexTermMapFilter` (scope by `:scope:` path prefix), and `GenIndexNodeBuilder` (build the node tree). `IndexCollectorPass` is now a thin orchestrator over these.

Kept this as a pure code-organization split rather than also adding an extension seam for other compiler passes to contribute/remove entries -- there's no second consumer yet to validate that shape against, so it'd be speculative. Happy to design that seam once there's an actual use case driving it.

packages/guides/resources/template/html/body/genindex/row.html.twig
linawolf: done

packages/guides/src/Nodes/Index/GenIndexRow.php
linawolf: Went with the smaller fix from the other thread instead: dropped isLink()/isSee()/isSeeAlso() and switch on getKind() directly in row.html.twig.

Reasoning: IndexCollectorPass (well, now IndexEntryCollector/GenIndexNodeBuilder after the split above) already uses the kind discriminator throughout for sorting and building rows before any node exists -- splitting GenIndexRow into GenIndexLink/GenIndexSee/GenIndexSeeAlso node types would remove the branch from the template, but the collector still needs a plain discriminator to do its sorting/building, so the win is narrower than "no more switches anywhere." Also touches quite a bit more surface (three node types, three templates, three renderer registrations, plus the collector's construction logic) for a template that's five lines. Open to revisiting if a real second consumer shows up that would benefit from the type split.

linawolf:
@jaapio I applied all requested changes, please resume review

linawolf:
@jaapio Replaced the phpstan-type array shapes in the IndexCollector classes with real value classes (GenIndexRowData, GenIndexTermData, GenIndexTermMap) per your Slack feedback and the article you linked. Ready for another look.

packages/guides-restructured-text/src/RestructuredText/Directives/GenIndexDirective.php
```diff
@@ -0,0 +1,66 @@
...
+ *   .. genindex::
+ *        :scope: Changelog/12.4/
+ *        :no-letter-index:
+ * ```
+ *
+ * The node is populated later by IndexCollectorPass, once entries from every
+ * document have been collected; at parse time it's an empty placeholder.
+ */
+#[Attributes\Directive(name: 'genindex')]
+final class GenIndexDirective extends BaseDirective
```
jaapio, line 56: the options `scope` and `no-letter-index` seem to be missing on this directive.

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
```diff
@@ -13,7 +13,153 @@
...
+ *
+ * .. index:: Backend, PHP-API, NotScanned, ext:core
+ *
+ * The directive itself is invisible in the rendered page; entries are
+ * collected project-wide to build the `genindex` page.
+ *
+ * @link https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-index
+ */
+#[Attributes\Directive(name: 'index', rawContent: true)]
+final class IndexDirective extends BaseDirective
```
jaapio, line 59: The `:name:` option described in the linked sphinx docs seem to be missing?

jaapio (commented):
I have the feeling that the index terms should not be part of the section node. It's more like some kind of metadata that we want to add to the nodes? What if we would add this as a more generic node element? 

If I see the rendering as a data element, and combine that with options I'm researching with https://pagefind.app/docs/metadata/ this could be very useful, 

Maybe it makes sense to add some metadata property to the abstractNode, That will make it more generic. 

```
public function addMetaData(DataNode $node)
```

The rendering should check the metadata for nodes it can handle. We could use a custom twig filter/function for this to render the data attributes on the nodes. 

```
<h1 data-index="{{ renderIndexData(node) }}">
```

This new system will help us to add attributes to the html elements in a more generic way, allowing people to do even more, like what I try to do with the pagefind, to make static search possible.

packages/guides-restructured-text/src/RestructuredText/Directives/GenIndexDirective.php
linawolf: Declared both in b6fe3482: `:scope:` and `:no-letter-index:` are now declared options and read through `readOption()`, so they show up wherever the directive's options are listed. A bare `:no-letter-index:` still turns the letter index off.

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
linawolf: Added in 1be87896: a named index block now registers a target, so :ref:`name` links to the section the block's entries are filed under, which is also where the index itself points.

packages/guides-restructured-text/src/RestructuredText/Directives/IndexDirective.php
linawolf: `$closestType` now starts out as the first `IndexEntryType` case, so it can never be null. Only the threshold check is left, and that's already a single early return.

linawolf:
@jaapio I rebased onto main and addressed the latest review:
- The index terms are no longer a property of `SectionNode`. `AbstractNode` has a generic `addMetaData(DataNode)`, and a `renderDataAttributes()` Twig function renders whatever metadata a node carries as `data-*` attributes (cd15b9d2). The index terms are its first user, and the rendered output is unchanged (`data-guides-index-terms`).
- `genindex` declares `:scope:` and `:no-letter-index:` (b6fe3482).
- `index` supports `:name:` as a `:ref:` target (1be87896).

Ready for another look.

PR #1362, opened 2026-08-26
Asset management

wouterj:
Symfony hosts their images in a different location on the server than the documentation articles. For this reason, we need to extract all images used in the docs and put them in a different folder (1). And we need to override the src path for images when rendering (2).  
This PR implements a solution for both of these problems, although (2) is the only blocker - (1) is just a nice-to-have (the current process simply assumes all assets are in `_images/` and copies that directory over).

The main thing is decoupling asset collection in a dedicated pre-renderer, instead of coupling it to the `asset()` Twig function. The prerenderer also allows specifying a different asset output filesystem. This allows us to immediately write it to the correct path. I can imagine this is also useful in more situations, e.g. when you want to immediately upload the assets to an S3-compatible bucket.
The `asset()` Twig function then is only concerned with generating a correct output path. I've added the option to have a different URL generator for images as well, so you could specify a different CDN host where your assets are published.

If the user doesn't change anything, the outcome will not change: images are stored in the destination file system, and the default URL generator is used to generate the public path.

The base uri can be configured in the CLI tool using a `assets_base_uri` option. I've chosen to not add a config for the asset destination. This feels a bit too niche, not worth the complexity of adding a config for it imho.

### Todo

* [x] Add functional test for image collection (I've verified it works manually)
* [x] Verify things work correctly when specifying a destination path.


tests/Integration/tests-full/md-to-rst/md-to-rst-list-link/expected/logs/error.log
wouterj, line 1: Benefit of decoupling image collection from the Twig function: this test references images that don't even exist. It didn't report this error before, as the md-to-rst templates don't parse image references.

packages/guides/tests/unit/Event/PostRenderDocumentTest.php
```diff
@@ -31,8 +33,8 @@ public function testEventCreation(): void
             RenderContext::forDocument(
                 $document,
                 [$document],
-                $this->createMock(FilesystemInterface::class),
-                $this->createMock(FilesystemInterface::class),
+                FlySystemAdapter::createFromFileSystem(new Filesystem(new InMemoryFilesystemAdapter())),
```
wouterj: As far as I could tell, the old mock was mocking something related to Flysystem V1, while the dependencies have Flysystem V3 installed. The new situation mocks the V3 situation.

tests/Integration/tests-full/md-to-rst/md-to-rst-list-link/expected/logs/error.log
jaapio: Nice!

jaapio:
Nice work! This has been on my wish-list for a while, but I never got to it.

wouterj:
Thanks! It has been on my todo list for a while now as well, glad to finally get some energy to contribute it. From my point of view, I believe this PR is now ready.

packages/guides/src/Twig/AssetsExtension.php
```diff
@@ -42,9 +38,7 @@ final class AssetsExtension extends AbstractExtension
 
     /** @param NodeRenderer<Node> $nodeRenderer */
     public function __construct(
-        private readonly LoggerInterface $logger,
```
jaapio, line 45: Removing these parameters would be a BC break. I would like to keep them for now. Maybe with a deprecation?

packages/guides/src/RenderContext.php
```diff
@@ -110,12 +115,14 @@ public static function forProject(
...
         string $destinationPath,
         string $ouputFormat,
+        FileSystem|null $imageDestination = null,
     ): self {
         $self = new self(
             $destinationPath,
             null,
             $origin,
             $destination,
+            $imageDestination ?? ($destination instanceof FilesystemInterface ? FlySystemAdapter::createFromFileSystem($destination) : $destination),
```
jaapio, line 125: I think you intended to use $imageDestination here?


```suggestion
            $imageDestination ?? ($imageDestination instanceof FilesystemInterface ? FlySystemAdapter::createFromFileSystem($destination) : $destination),
```

packages/guides/tests/unit/Event/PostRenderDocumentTest.php
```diff
@@ -31,8 +31,8 @@ public function testEventCreation(): void
             RenderContext::forDocument(
                 $document,
                 [$document],
-                $this->createMock(FilesystemInterface::class),
-                $this->createMock(FilesystemInterface::class),
+                FlySystemAdapter::createInMemory(),
```
jaapio, line 34: ❤️

jaapio (commented):
This looks good. If you are able to fix the few remarks I had this is good to merge.

packages/guides/src/RenderContext.php
wouterj: I was confused as well, but the existing code is correct I think:

We either use `$imageDestination` when given (left side of `??` operator), which already always is a `FileSystem` instance (no need for BC on a new parameter).
Or we default to `$destination` (right side of `??` operator). In that case, we have to make sure to convert the old-signature `Flysystem\FilesystemInterface` to a `FileSystem` instance.

jaapio (approved):

PR #1364, opened 2026-08-28
[FEATURE] Support selecting an alternate template per document

linawolf:
Add the generic `:template:` field-list metadata: TemplateNode,
TemplateFieldListItemRule to parse it, TemplateMetadataNodeRenderer to
render it as nothing (it's a marker, not visible content), and
DocumentNode::getTemplate() to expose the selected template name.

Split out of #1358 per review feedback -- the genindex work on that
branch is the first consumer (`:template: genindex`), but the
mechanism itself is generic and reviewable on its own.

Signed-off-by: lina.wolf <lwolf@w-commerce.de>

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01PP4LkejR5PSubbhNmF4RkT)_

### DIFF
--- a/packages/guides-restructured-text/resources/config/guides-restructured-text.php
+++ b/packages/guides-restructured-text/resources/config/guides-restructured-text.php
@@ -89,6 +89,7 @@ use phpDocumentor\Guides\RestructuredText\Parser\Productions\FieldList\Organizat
 use phpDocumentor\Guides\RestructuredText\Parser\Productions\FieldList\OrphanFieldListItemRule;
 use phpDocumentor\Guides\RestructuredText\Parser\Productions\FieldList\ProjectFieldListItemRule;
 use phpDocumentor\Guides\RestructuredText\Parser\Productions\FieldList\RevisionFieldListItemRule;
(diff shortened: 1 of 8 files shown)


jaapio (commented):
Could you also include an integration test for this new feature?

linawolf:
No integration test here: at this point `:template:` only reaches `DocumentNode::getTemplate()` — nothing consumes it yet, and `TemplateMetadataNodeRenderer` always renders it as an empty string regardless of the value. There's no observable output difference to assert end-to-end, so an integration test isn't possible for this PR in isolation.

The unit tests cover what's actually testable at this layer: `TemplateFieldListItemRuleTest` (parses `:template:` case-insensitively into a `TemplateNode`) and `TemplateMetadataNodeRendererTest` (only `TemplateNode` renders as nothing) — same pattern as the other marker fields (`orphan`, `navigation-title`).

An integration/fixture test showing the template actually affecting output belongs with the consumer — the genindex work on the follow-up branch, which reads `getTemplate() === 'genindex'`.

jaapio (approved):

PR #1371, opened 2026-08-31
Support PHPUnit 12 alongside 10.5

linawolf:
Continues #1327.

Drop CoversMethod attributes incompatible with locked PHPUnit 10.5

CoversMethod is a PHPUnit 11/12-only attribute, but composer.lock still
resolves phpunit/phpunit to the 10.5 line under the new "^10.5 || ^12.0"
constraint, so PHPStan (and PHPUnit itself at runtime) can't find the
attribute class. CoversClass already attributes coverage to the whole
class, so nothing is lost by dropping the redundant per-method attributes.

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: lina.wolf


linawolf:
@jaapio this continues your work from #1327 — rebased onto current main, plus a fix for a PHPUnit 10.5 compatibility issue in the CoversMethod attributes that was failing static analysis there. Would you mind taking a look?

jaapio (approved):
Thanks a lot! I'm waiting for another pr to be merged. It's very likely this will cause a conflict as the composer.lock is updated in both.

PR #1372, opened 2026-08-31
[BUGFIX] Resolve cross-page fragment links in markdown

linawolf:
Markdown links combining a page reference and a fragment anchor, e.g.
`[text](page-b.md#section-two)`, failed to resolve, logging "Reference
page-b.md#section-two could not be resolved". Two independent bugs caused
this:

1. `LinkParser` checked `str_ends_with($url, '.md')` on the full URL
   including the fragment, so the `.md` extension was never stripped.
2. `PageHyperlinkResolver` passed the full target reference (including
   `#fragment`) to `findDocumentEntry()`, which never matches since no
   document is keyed with a fragment suffix.

This ports the fix from phpDocumentor/guides#1300 (originally authored by
Jonathan Hefner <jonathan@hefner.pro>, open since 2026-02-07 and stalled
on a stale-branch/CI mixup unrelated to the fix itself), rebased onto
current `main`, plus additional regression test coverage:

- An integration-test link with two `#` characters
  (`page-b.md#section-two#extra`), confirming only the first `#` is
  treated as the fragment delimiter and the rest is preserved verbatim.
- A matching `PageHyperlinkResolver` unit test case for the same
  multi-hash scenario.
- A new `LinkParserTest`, with data-provider cases covering: no anchor,
  an anchor (the primary bug), a duplicate/multi-hash anchor, an
  anchor-only link with no page, and a fragment that itself ends in
  ".md" (a regression guard against `str_ends_with()` matching on the
  wrong part of the string).

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: lina.wolf

### DIFF
--- a/packages/guides-markdown/src/Markdown/Parsers/InlineParsers/LinkParser.php
+++ b/packages/guides-markdown/src/Markdown/Parsers/InlineParsers/LinkParser.php
@@ -22,7 +22,9 @@ use phpDocumentor\Guides\Nodes\Inline\PlainTextInlineNode;
 use Psr\Log\LoggerInterface;
 
 use function assert;
+use function explode;
 use function filter_var;
+use function str_contains;
 use function str_ends_with;
 use function substr;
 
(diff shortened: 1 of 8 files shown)


jaapio (approved):

jaapio:
Thanks @linawolf and @jonathanhefner for this fix!

Issue #1378, opened 2026-09-02
Migrating a directive to #[Directive]/createNode() can silently break parse-time-dependent or ordering-dependent behavior


linawolf:
Follow-up from #1373. Three directives that looked like trivial migrations broke real behavior once upgraded. Root cause is structural, so filing separately -- it'll keep hitting the #1373 effort otherwise.

Once a directive has #[Directive], DirectiveRule::apply() stops calling process()/processSub() at parse time -- createNode() only runs later, from DirectiveProcessPass at compile time, after every document is already fully parsed. That breaks anything assuming the node existed earlier.

**Confirmed broken by migrating and running the test suite:**
- ClassDirective -- broke `class-in-list`/`big-numbers` integration tests (ClassNodeTransformer, priority 40,000, runs before DirectiveProcessPass, priority 100, so the class-marker node doesn't exist yet when the transformer tries to apply it to the next sibling)
- TabDirective -- broke `tabs_html` functional test (rendered completely empty): parent TabsDirective is still old-style and inspects its children synchronously at parse time, but migrated TabDirective's children are still unresolved placeholders then
- DocumentBlockDirective -- broke 13 bootstrap/menu integration tests: whatever collects its content into parts.footer for the theme template runs before DirectiveProcessPass resolves it, so the footer ends up empty and the content leaks into the main body

**Flagged by code inspection, not yet migrated/tested:**
- TestLoggerDirective, SectionauthorDirective, ConfigurationBlockDirective, TabsDirective -- all log warnings via $blockContext->getLoggerInformation() (file/line), which createNode() has no equivalent access to
- RawDirective -- needs literal unparsed source text, but the compile-time content parser always fully RST-parses first

**Proposed fix for the logging case:** snapshot the needed info into Directive/DirectiveNode at parse time instead of trying to carry BlockContext across the phase boundary. No fix proposed yet for the raw-content or transformer-ordering cases.

Any directive with logging, raw-content needs, or a role in another pass's ordering needs case-by-case review before migrating, not just "add the attribute and go."

linawolf:
Two more ordering hazards, found migrating another batch (Bootstrap cards/accordions, list-table). Neither is the BlockContext/logging/raw-content shape already covered here.

**Confirmed broken by migrating and running the test suite:**

[...]

Issue #1373, opened 2026-08-31
Deprecated directive without attributes


jaapio:
Directives must contain a `Directive` attribute and when applicable `Option` attributes. 

```
#[Directive(name: 'example')]
#[Option(name: 'option1', type: OptionType::Boolean, description: 'An example option', default: false)]
class ExampleSubDirective extends SubDirective
{
    public function createNode(\phpDocumentor\Guides\RestructuredText\Nodes\DirectiveNode $directiveNode): Node
    {
        return new ExampleNode(
            $this->readOption($directiveNode, 'option1'),
            $directiveNode->getDataNode(),
            $this->text,
            $directiveNode->getChildren(),
        );
    }
}
```
 
The new way of defining directives allows us to do validation on the options set on the `DirectiveNode`. But also brings us closer to the options to do validation on the parsed document. Directives migrated to this new setup MUST implement the `createNode` method, which is responsible for the creation of the final node used during rendering. 

## Steps to migrate:

- Move the logic to create the node(s) created by the directive into the createNode method. 
- Remove the `getName` and `getAliases` method and replace them with the `Directive` attribute
- Remove the `processSub` and `process` methods

## More info

Read more about directive in our docs: https://docs.phpdoc.org/components/guides/guides/developers/directive.html#implement-a-sub-directive

See: https://github.com/phpDocumentor/guides/pull/1231

PR #1380, opened 2026-09-02
[FEATURE] Add synchronous dispatch escape hatch for #[Directive]

linawolf:
[FEATURE] Add synchronous dispatch escape hatch for #[Directive]

Migrating a directive to the compile-time createNode() model defers
node creation from parse time to compile time -- createNode() has no
BlockContext, so any directive whose logic depends on live parser
state (logging with line numbers, resolving the current file's path,
raw unparsed content, or re-entering the parser for nested content)
silently loses that ability if migrated the normal way. Found via
#1378 while migrating directives for #1373.

#[Directive] gains an optional synchronous flag: when set, the
directive stays fully declared (name/aliases/options visible to
reflection-based tooling like jaapio's reference-doc generator) but
keeps running through the original process()/processSub() path with
a real, live BlockContext, instead of deferring to createNode() at
compile time.

Proves the concept on the two hardest cases found in #1378:
TestLoggerDirective (asserts an exact log line including parse-time
line number) and RawDirective (needs literal unparsed source text).
Neither needed any change to their actual logic -- just the one-line
attribute addition.

Verified via Docker (php:8.4-cli, matching CI): phpstan clean, phpcs
clean, deptrac 0 violations, unit/functional/integration suites green
(970 tests, 0 failures) -- including byte-for-byte confirmation that
TestLoggerDirective's log line and RawDirective's raw-content tests
are unchanged.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: lina.wolf

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Directives/Attributes/Directive.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Directives/Attributes/Directive.php
@@ -18,10 +18,20 @@ use Attribute;
 #[Attribute(Attribute::TARGET_CLASS)]
 final class Directive
 {
-    /** @param string[] $aliases */
+    /**
+     * @param string[] $aliases
+     * @param bool $synchronous Keep processing this directive synchronously at parse
+     *     time via process()/processSub(), instead of deferring node creation to
+     *     createNode() at compile time. Declaring name/aliases here still makes the
+     *     directive discoverable by tooling that reads this attribute; set this when
+     *     the directive's logic genuinely depends on parse-time-only state (e.g. the
+     *     current BlockContext for logging, file paths, or raw/unparsed content) that
+     *     createNode() has no access to.
+     */
     public function __construct(
         public readonly string $name,
         public readonly array $aliases = [],
+        public readonly bool $synchronous = false,
     ) {
     }
 }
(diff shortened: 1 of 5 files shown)


jaapio:
I think this breaks the concept of having the directives compile time. If we need more info we should add it into the `DirectiveNode` the main reason to introduce this concept of having compile time directives was to be able to stick closer to the orginal document, so we can introduce validation on the AST. Which would be one of the next things I would like to build in.

linawolf:
Closing unmerged per the discussion above. Keeping `synchronous: true` as a silent, sanctioned bypass would work against the eventual v2 removal of the old process()/processSub() dispatch path, and doesn't actually buy anything the existing "no #[Directive] attribute yet" state doesn't already give us -- that state already fires a deprecation and supports migrating in small steps.

For the record, these deprecations are silent by default for real consumers of the library (doctrine/deprecations only surfaces them if something explicitly opts in, e.g. our test bootstrap) -- so leaving directives unmigrated in the meantime doesn't put end users at risk.

Will follow up on the real per-directive fixes (extending DirectiveNode with source location for the logging/path cases, and figuring out a compile-time-compatible answer for the directives that currently re-enter the parser synchronously) separately.

PR #1381, opened 2026-09-02
[TASK] Migrate remaining directives to #[Directive] via synchronous dispatch

linawolf:
[TASK] Migrate remaining directives to #[Directive] via synchronous dispatch

The remaining directives in guides-restructured-text/Directives/ (or
their external consumers, for Class/Tab/DocumentBlock/Option) all
depend on live BlockContext for logging, path resolution, raw content,
or a role in another pass's ordering -- so they use the synchronous
escape hatch from #1380 rather than the deferred createNode() model,
per #1378's findings. Their process()/processSub() logic is untouched.

Leaves only GeneralDirective (catch-all fallback) and ActionDirective
(abstract base) in this package without the attribute.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: lina.wolf

### DIFF
--- a/packages/guides-restructured-text/src/RestructuredText/Directives/ClassDirective.php
+++ b/packages/guides-restructured-text/src/RestructuredText/Directives/ClassDirective.php
@@ -25,25 +25,14 @@ use function array_map;
 use function array_merge;
 use function explode;
 
+/**
+ * When the default domain contains a class directive, this directive will be shadowed. Therefore, Sphinx re-exports it as rst-class.
+ *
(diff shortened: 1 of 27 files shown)


linawolf:
Closing unmerged. This was built entirely on the `synchronous: true` mechanism from #1380, which we're not pursuing (see that PR's discussion) -- it would make the old process()/processSub() dispatch permanent for anything that uses it, undermining the eventual v2 removal of that path, without actually gaining anything the existing "unmigrated, deprecation-warned" state doesn't already provide.

The directives here stay unmigrated for now (still firing the existing #1373 deprecation) until each one's real blocker is solved individually.

PR #1393, opened 2026-09-07
[FEATURE] Support text language and direction (lang/dir) in ReST

### LINKED ISSUE #1266: Ability to specify text direction in ReST
### Feature request

It would be great to be able to specify text direction in sections of text. 

This could for example be achieved with `..  role:: rtl` (for blocks). Maybe an inline option should also be possible? For example ``Example in Arabic is :rtl:`مثال` ``

### LINKED ISSUE #1267: Ability to specify text language in ReST
### Feature request

It would be great to be able to specify text language for blocks of text in ReST.

This could for example be achieved with `.. role:: language(nb)` to set the language to Norwegian Bokmål (ISO code `nb`). An inline version of this would also be great. Maybe ``Example in Norwegian is :language(nb):`eksempel` ``

linawolf:
Resolves #1266
Resolves #1267

Authors had no way to mark a block, a run of inline text, or a whole
page as being in a different language or text direction, so the
rendered HTML never got the lang/dir attributes needed for correct
rendering (e.g. RTL scripts, screen readers, correct font/glyph
shaping). The maintainers asked what concrete output this should
produce.

docutils/Sphinx were checked for prior art before designing this:
- Document level: neither has any per-document mechanism at all (only
  project-wide build config), so nothing to align with there.
- Inline level: docutils' own FAQ recommends exactly `:rtl:`/`:ltr:`
  roles for this, just via `.. role::` boilerplate the author has to
  write themselves; this makes them built-in instead.
- Block level: docutils has `.. class:: language-<tag>` and
  `.. class:: rtl`/`ltr`, but that's CSS-only styling hooks with no
  real HTML lang/dir attribute, so it doesn't actually address the
  accessibility problem these issues are about. Kept our own
  `:lang:`/`:dir:` container options, which do emit real attributes.

This reuses existing mechanisms at three levels rather than inventing
new directives:

- Block-level: `:lang:`/`:dir:` options on the existing `.. container::`
  / `.. div::` directive.
- Inline-level: new `:rtl:`/`:ltr:` roles for pure direction switches,
  and a combined `:lang:` role (`:lang:`text (language, direction)`)
  for text tagged with both, following the same trailing-`(...)`
  convention as the existing `:abbreviation:` role.
- Document-level: new `:lang:`/`:dir:` field-list metadata, mirroring
  the existing `:template:` field, setting the page's own `<html>` tag.

`dir` is validated against ltr/rtl/auto and `lang` against a loose
BCP 47 shape; both are warn-only diagnostics, never blocking, matching
the ValueType validation philosophy already used elsewhere.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PP4LkejR5PSubbhNmF4RkT


packages/guides-restructured-text/src/RestructuredText/Directives/ContainerDirective.php
```diff
@@ -14,20 +14,83 @@
...
+ *     Some content in Arabic.
+ *
  * @link https://docutils.sourceforge.io/docs/ref/rst/directives.html#container
  */
 #[Attributes\Directive(name: 'container', aliases: ['div'])]
+#[Option(name: 'lang', description: 'Sets the HTML lang attribute on the wrapping div, e.g. "en", "nb", "ar".')]
+#[Option(name: 'dir', description: 'Sets the HTML dir attribute on the wrapping div: "ltr", "rtl", or "auto".')]
 final class ContainerDirective extends SubDirective
 {
+    private const VALID_DIRECTIONS = ['ltr', 'rtl', 'auto'];
```
jaapio, line 46: Feels like this is a good candidate for an enum?

packages/guides-restructured-text/src/RestructuredText/TextRoles/LangTextRole.php
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * ```rest
+ * :lang:`مثال عربي (ar)`
+ * :lang:`مثال عربي (ar, rtl)`
+ * ```
+ */
+final class LangTextRole extends BaseTextRole
+{
+    private const VALID_DIRECTIONS = ['ltr', 'rtl', 'auto'];
+    private const LANGUAGE_TAG_PATTERN = '/^[a-zA-Z]{2,8}(-[a-zA-Z0-9]{1,8})*$/';
```
jaapio, line 40: This seems to be repeated multiple times. Maybe we can use an interface with a constant to specify this? Something like `Language` 

```
interface Language 
  const PATTERN = 
```

jaapio (changes requested):
I think direction should be an enum rather than an arbitrary string, this allows us to be more strict. We can validate the value before we create the enum, and otherwise fall back to `auto` when the value is invalid. 

This will make sure rendering always works even if the userinput is wrong.

packages/guides-restructured-text/src/RestructuredText/Directives/ContainerDirective.php
linawolf: Done

packages/guides-restructured-text/src/RestructuredText/TextRoles/LangTextRole.php
linawolf: Done

linawolf:
@jaapio changes applied, please have another look

PR #1395, opened 2026-09-17
Fall back to the target reference when a label has no title

### LINKED ISSUE #1394: A :ref: to a label outside a section renders an empty link
`AnchorReferenceResolver` fills a reference that has no children with `$target->getTitle() ?? ''`. A target only carries a title when it is a section, or an anchor whose parent in the shadow tree is a `SectionNode` — `CollectLinkTargetsTransformer` sets it from `$parentSection->getTitle()` and leaves it `null` otherwise. A label declared anywhere else therefore resolves to a link with no text at all.

Minimal reproduction:

```rst
..  note::

    .. _nested-label:

    A label declared inside a directive body.

A plain reference to it: :ref:`nested-label`
```

Rendered with this repository's own default templates at `c9868443`:

```html
<p>A plain reference to it: <a href="/index.html#nested-label"></a></p>
```

A resolved link, a correct `href`, and nothing between the tags. Nothing is logged, because from the resolver's point of view the reference resolved fine. The same page with the label on a section renders `<a href="/index.html#section-label">A section with a label</a>`, so it is specifically the titleless target that produces the empty link.

Two consequences. For a reader the link is invisible; for a screen reader it has no accessible name at all, which fails WCAG 2.4.4 *Link Purpose (In Context)* and 4.1.2 *Name, Role, Value*, both Level A. And since it happens only for labels outside a section, it tends to appear in exactly the places an author reached for a label deliberately.

`InterlinkReferenceResolver` has the same shape but cannot hit it: `InventoryLink::$title` is a non-nullable `string`.

The fix I would suggest is to fall back to the reference the author wrote rather than to the empty string. PR: #1395.

CybotTM:
Fixes #1394.

`AnchorReferenceResolver` fills a childless reference with `$target->getTitle() ?? ''`, and a target only carries a title when it is a section or an anchor whose parent in the shadow tree is one — `CollectLinkTargetsTransformer` leaves it `null` everywhere else. A label declared inside a directive body therefore resolved to a link with a correct `href` and nothing between the tags: invisible to a reader, without an accessible name for a screen reader, and nothing logged, because from the resolver's side the reference resolved.

This falls back to the reference the author wrote instead of the empty string. One line, plus a comment saying why the title can be absent, since the reason is two classes away.

## Measured, before and after

The same source rendered with this repository's own default templates, at `c9868443` and with the patch applied:

```
before   <p>Reference to the label:   <a href="/index.html#label-without-section"></a></p>
after    <p>Reference to the label:   <a href="/index.html#label-without-section">label-without-section</a></p>

before   <p>Reference to the section: <a href="/index.html#label-on-a-section">A section carrying a label</a></p>
after    <p>Reference to the section: <a href="/index.html#label-on-a-section">A section carrying a label</a></p>
```

The second pair is the point: a target that has a title keeps it.

## Why the author's reference rather than the anchor

`$node->getTargetReference()` is what stands in the source; `$target->getAnchor()` is the normalized form. For the case at hand they are usually the same string, and where they differ the author's spelling is the one they can search for.

## Tests

`tests/Integration/tests/anchor/anchor-without-title` renders both cases in one file. Reverting only the resolver line makes it fail on exactly the titleless link and leaves the titled one untouched, so the fixture pins the rendered behaviour rather than merely visiting it.

Two unit cases in `AnchorReferenceResolverTest`, both asserting that nothing is logged, as the two existing cases in that file do:

- `testTitledTargetIsUsedAsLinkText` — a titled target still uses its title. This is the guard: making the fallback unconditional fails it, which I checked rather than assumed.
- `testTargetWithoutTitleFallsBackToTheTargetReference` — the case above; it fails with `''` against the unchanged resolver.

`phpunit` 980 tests, `make code-style` and `make phpstan` clean.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NxSeVq1hDnGBCqKLGcjQ6m)_


CybotTM:
Added since the first push: an integration fixture, `tests/Integration/tests/anchor/anchor-without-title`. The unit cases pin the resolver, but nothing pinned the rendered output, and the empty link is only visible there. Reverting the one resolver line makes the fixture fail on exactly the titleless link and leaves the titled one untouched, so it distinguishes the defect rather than merely visiting the code path. The two unit cases now also assert that nothing is logged, as the two existing ones in that file do.

I also corrected the issue text: it quoted the rendering as `<a href="#nested-label"></a>`, while this repository's default templates emit `<a href="/index.html#nested-label"></a>`. Same defect, but a reader reproducing it would have seen a different string.

55 checks green, `phpunit` 980, `make code-style` and `make phpstan` clean.

linawolf (approved):

linawolf (approved):
Thanks, this fixes the empty link text nicely. An untitled anchor now shows the reference exactly as the author wrote it, and titled targets still use their section heading. The new `anchor-without-title` fixture plus the unit tests cover both cases.

Optional follow-up, not blocking: `AnchorHyperlinkResolver` (line 62) and `TitleReferenceResolver` (line 55) still fall back to an empty string. In practice neither is likely to hit that: hyperlinks usually have their own text, and title targets always have a title. Switching them to the same fallback would still keep all three resolvers consistent.

LGTM 👍

PR #1397, opened 2026-09-18
[TASK] Restore TestExtension in the shipped source and document it

CybotTM:
Merging this puts `phpDocumentor\Guides\DependencyInjection\TestExtension` back into `packages/guides/src/`, so it ships with `phpdocumentor/guides` again and downstream test suites that construct it keep working. It also adds a documentation chapter and a class docblock stating that the class is a supported testing utility, what it changes in a container, and what a container has to provide before it can run.

## Why

#1357 moved the class to `packages/guides/tests/unit/`, which every package marks `export-ignore`, so it stopped being part of the published package. @wouterj [pointed out](https://github.com/phpDocumentor/guides/pull/1357#issuecomment-5729073786) that this is a BC break: `wouterj/symfony-docs-guides` builds its test container with `DocsKernel::create([new TestExtension()])` in `lib/symfony-extension/tests/integration/HtmlIntegrationTest.php` and asserts on the `TestHandler` the extension registers. A code search across GitHub finds that repository and this one as the only two source consumers, but "only one external consumer today" is not an argument for removing a public class from a released package.

The reason given for the move was that the class references `Monolog\Logger`, `Monolog\Handler\TestHandler` and `Symfony\Component\Clock\MockClock`, "none of which the package requires". That was wrong on its face: `symfony/clock` has been a direct requirement of `packages/guides/composer.json` since `fd5694c7` (2023-11-22), so `MockClock` is covered and only Monolog is undeclared. And an undeclared Monolog is a class-string reference that resolves without the class existing. They matter only when a consumer actually registers `TestExtension`, at which point the consumer is in a test environment that has Monolog. I no longer think that cost justifies taking a working test helper away from downstream packages.

The honest trade-off: this reopens one item of #1355 — the "Test support in production `src`" paragraph, which names #1357 as its resolution. I'd now resolve that one item as won't-fix and leave the rest of #1355 standing; the undeclared Monolog can be recorded as a `suggest` or simply accepted.

## Timing

The break is not in any release. `2f152ae8` is contained in no tag — the newest is [1.10.2](https://github.com/phpDocumentor/guides/releases/tag/1.10.2) from 2026-05-12, and there is no `1.x` branch for `.backportrc.json` to target — so it exists only on `main`. Merging this before the next tag means no published version of `phpdocumentor/guides` was ever without the class, and no backport is needed. It also means the window for deciding this closes at the next release.

## What the documentation says

`docs/developers/extensions/testing.rst`, added to the extension chapter's toctree, covers the three things `process()` does — making `CommandBus`, `Parser`, `Compiler` and `phpdoc.guides.output_node_renderer` public; defining `ClockInterface` as a `MockClock` at `2023-01-01 12:00:00`, overriding the `NativeClock` alias `guides-cli` registers, which is what stamps the `ProjectNode` date a fixture compares against; and registering `TestHandler` and pushing it onto the `Logger` — plus a worked example of registering it from a `TestCase` and how to assert on log records.

It also names the prerequisite that bites first: the `Monolog\Logger` service comes from `ApplicationExtension` in `phpdocumentor/guides-cli`, not from `phpdocumentor/guides`, so a container built without it fails in `process()` with a `ServiceNotFoundException`. A package depending on `phpdocumentor/guides` alone needs `monolog/monolog` in its own `require-dev`. The class docblock carries the same facts for anyone who reaches the class before the docs.

## Not done here

No `suggest` entry for `monolog/monolog` was added to `packages/guides/composer.json`. It would document the soft dependency in a machine-readable place, but it touches package metadata that the sub-split publishes, so it seemed better raised than slipped in — say the word and I will add it.

## Verification

Rebased onto `main` at `6a547e5e` and re-run on that tree, PHP 8.5.10: PHPUnit 990 tests / 6449 assertions, exit 0. PHPCS, PHPStan and deptrac (`deptrac.packages.yaml`) all exit 0. The documentation gate from `.github/workflows/documentation.yml` — `vendor/bin/guides -vvv --no-progress docs --fail-on-log` — exits 0 with the new page rendered and linked from the extensions index. The `xml-lint` job is the one gate I did not reproduce locally (no `libxml2-utils` on this machine); the diff contains no XML.

Draft, because the first commit reverses a decision that was already merged and the second one commits the project to supporting this class: both are yours to confirm rather than mine.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01EWiYVcHSD71piDkNv2r8c6)_


CybotTM:
Sorry fo #1357 - provided this PR as a revert and enhancement.

jaapio (approved):

